PR #6, opened 2026-05-04
tui: order child graph rows by blockers

mariusvniekerk:
## Summary
- hydrate outgoing `blocks` metadata onto list issue rows so expanded child graph rows can be dependency-aware
- default expanded child rows to topological blocker order, with `o` toggling between topological and temporal graph order
- update TUI help/footer snapshots to expose `o order` and graph terminology

## Note
The visible terminology now says graph, but several internal TUI helpers and types still use inherited `queue*` names (`queueRow`, `buildQueueRows`, `queueHelpRows`, working-set cache names). Those are now domain-misaligned and may deserve a focused rename pass, but I left them alone here to keep this PR about behavior plus user-facing wording rather than a broad mechanical churn.

## Validation
- `go test ./internal/tui`
- `go test ./internal/daemon`
- `go test ./internal/db`
- `GOFLAGS=-buildvcs=false go test ./...`

Plain `go test ./...` hit Go VCS stamping failures in nested test builds (`error obtaining VCS status`), so the full-suite verification used `GOFLAGS=-buildvcs=false`.


wesm:
Not valid. merging

PR #8, opened 2026-05-04
feat(daemon): opt-in remote-client mode (no auth)

cwensel:
## Summary

Adds opt-in remote-client mode so a kata daemon can serve clients on other hosts over a private network (tailscale, VPN, etc.) without auth — network ACLs are the boundary.

- Daemon: new `--listen host:port` flag and `<KATA_HOME>/config.toml` `listen = "..."` for persistent TCP bind. Validation rejects public IPs, hostnames, and the unspecified bind; accepts loopback, RFC1918, CGNAT, link-local, and ULA.
- Client: `KATA_SERVER` env and `.kata.local.toml` `[server].url` resolve a remote daemon. `kata init` appends `.kata.local.toml` to `.gitignore`. A configured-but-unreachable remote returns `ErrRemoteUnavailable` (exit 7) instead of silently auto-starting a local daemon.
- Daemon resolves projects path-free when the client sends `project_identity` from its local `.kata.toml` — no filesystem stat on the daemon's side for paths that don't exist on the daemon's host.
- Specs: `docs/superpowers/specs/2026-05-04-kata-remote-client-design.md`, plan in `docs/superpowers/plans/`. README + CLAUDE.md updated.

## Test plan

- [x] `go test ./...` (unit + integration)
- [x] `e2e/remote_client_test.go`: daemon on TCP loopback under one `KATA_HOME`, client subprocess under a separate `KATA_HOME`/workspace via `KATA_SERVER`; init/create/list/close end-to-end; asserts client wrote no runtime files locally
- [x] roborev pass on every commit (Medium findings fixed; remaining Lows triaged)
- [ ] Manual smoke: tailnet-reachable host, `kata daemon start --listen <tailnet-ip>:7777`, second host with `KATA_SERVER` set


wesm:
I have a bunch of pending work that I'm going to merge to main, but once you hit a stopping point I can take care of rebasing your branch (there shouldn't be significant conflicts)

cwensel:
i'll be happy to try the rebase and drop out if i see any conflicts. i'd like this to bake a bit longer plus get feedback. 

it's behavior should be intuitive, also I went for least controversial naming as a could.

wesm:
roger that. I just merged my slug of pending work, so I would just have your agents rebase now

cwensel:
rebased. easy conflict. so far working fine.

wesm:
@cwensel you can add stuff like roborev has to review_guidelines in .roborev.toml to squelch some of the threat model stuff (it won't get picked up until it's merged into main of course) for next go around. I usually just say "this is only for use on secure networks for now"

cwensel:
Round 2 on the roborev findings, pushed in `a552ada`. Five fixes plus three I'd like to push back on:

## Fixed

- **Runtime TCP parsing** — `ParseAddress` now returns `TCPEndpointAny` for TCP inputs. The runtime file is daemon-written under `$KATA_HOME` 0700, so a same-host client must be able to dial back a non-loopback private `--listen` address. Strict loopback-only was correct for untrusted input but wrong for the trusted readback path; spec §4.1 updated. (`internal/daemon/endpoint.go`)
- **Subdir project resolution** — new `config.FindProjectConfig` walks upward for `.kata.toml`. Both `cmd/kata/create.go:resolveProjectID` and `internal/tui/client.go:ResolveProject` use it, so a subdir invocation sends `project_identity` (path-free) instead of falling back to a `start_path` the remote daemon can't stat. Bootstrap (`kata init`) keeps the exact-dir read since `.kata.toml` doesn't exist yet.
- **Nested-init `.gitignore`** — CLI now reads the daemon's `workspace_root` from the init response and places `.gitignore` there, so a `kata init` from a subdir lands the entry beside `.kata.toml` at the workspace root. Falls back to `startPath` only if `workspace_root` is missing (older daemon).
- **Workspace-scoped `.kata.local.toml`** — added `EnsureRunningInWorkspace(ctx, workspaceStart)`; CLI plumbs the resolved `--workspace` absolute path through. Running `kata --workspace /repo create` from outside the repo now picks up `/repo/.kata.local.toml`'s `[server]` override instead of falling through to local auto-start. TUI keeps the CWD-anchored variant.
- **`chooseEndpoint` TOCTOU** — replaced listen-then-close with the non-binding `daemon.ValidateNonPublicAddress`. Eliminates the window where the validating bind could lose the OS-allocated port (with `:0`) or race with another process for a fixed port. The actual bind happens once inside `server.Run`.

Tests added: `TestParseAddress_TCPAcceptsNonLoopbackPrivate`, `TestFindProjectConfig_*`, `TestInit_GitignoreLandsAtWorkspaceRoot`, `TestResolveRemote_WorkspaceAnchorOverridesCwd`, `TestResolveRemote_EmptyWorkspaceFallsBackToCwd`. Full suite passes.

## Pushed back

- **Unauthenticated TCP API** — out of scope per spec §1.4 / §2 (auth is an explicit non-goal: "network-level ACLs are the access boundary in v1") and §10.1 names shared-token auth as a follow-up. The PR title says "no auth" and the design doc backs it up. Adding auth here would be a much larger surface change; the next plan file is the right place.
- **`kata init` uses client paths on the server** (only one of three reviews flagged this) — `kata init` is the bootstrap call that runs *before* `.kata.toml` exists, so there's no identity to send and `start_path` is the only handle the daemon has. The path-free `project_identity` resolution kicks in for every command *after* init. The other two review passes did not flag this, consistent with the spec design.
- **Tracked `.kata.local.toml` redirects to attacker server** — `kata init` already adds `.kata.local.toml` to `.gitignore` so it stays untracked under normal use. Hardening with a `git ls-files` check (warn or refuse on a tracked file) would be a small follow-up; happy to do it as a separate change if you want it in this PR rather than later.

Spec updated for the runtime-parse, workspace-anchor, gitignore-landing-site, and no-bind-validate changes.

cwensel:
i've tested the remoteness most of the day with a mixed mode topology

host a: daemon w/ listen + kata tui (daemon auto starts - client auto discovers port etc)
host b: kata tui w/ remote url (no daemon launches)

wesm:
Giving this a quick look but I trust that it works, the project is so new that if there's a bug somewhere I'm sure we can patch it easily

wesm:
Path-free `kata init` and four roborev cycles since e250214 — summary for reviewers.

## What `kata init` does now

When the client can derive the project identity locally — from `.kata.toml`, `--project`, or a discoverable git workspace — it sends `project_identity` to the daemon and writes `.kata.toml` and `.gitignore` itself. The daemon never stats the client's filesystem. Falls back to the path-based request only when no identity source is available locally, so the daemon (or its absence) emits the existing validation error.

This makes `kata init` work the same way `kata create` and `kata list` already do under remote-client mode: the contract is identity-driven, not path-driven.

## What the reviews caught

**Cycle 1 — alias and identity semantics (aa3d69d)**

1. `--reassign` was accepted but ignored in path-free mode — user got success, but no alias moved.
2. Client-side conflict error dropped its `code` field, so `--json` consumers couldn't branch on `project_binding_conflict`.
3. Path-free init bypassed alias attach and conflict detection for all clients (including local), losing the rule that two projects can't bind to the same workspace alias.
4. `upsertProject` resolved through `project_aliases`: a path-free request whose identity matched an alias for project Y silently rebound to Y.
5. `kata health` ignored `KATA_SERVER` — it scanned local runtime files even when the user had explicitly selected a remote daemon.

Fix: client now computes alias metadata locally and sends it alongside `project_identity`; daemon does strict identity lookup and runs the existing alias-attach + reassign logic. `discoverDaemon` now consults the same remote-resolution path as `ensureDaemon`.

**Cycle 2 — paths with spaces (a6af41e)**

`alias.identity` was being run through `ValidateIdentity`, which forbids whitespace. Local aliases are `local://<absolute path>`, so a workspace at `/Users/me/My Project` produced `local:///Users/me/My Project` and got rejected — even though the path-based init flow attaches that exact alias without complaint.

Fix: new `ValidateAliasInfo` with kind-aware rules. Git aliases keep the strict charset; local aliases just need a `local://` prefix and a non-empty path. Side benefit: invalid `kind` and empty `root_path` now return a clean 400 instead of a 500 from a downstream DB CHECK constraint.

**Cycle 3 — masked .kata.toml parse errors (e5d8735)**

`resolveProjectID` (CLI) and `Client.ResolveProject` (TUI) both swallowed every error from `FindProjectConfig` and fell back to `start_path`. With a malformed `.kata.toml`, the request reached the daemon as a path resolution — and under remote-client mode that surfaces as `stat: no such file`, hiding the real fix-it error.

Fix: distinguish `ErrProjectConfigMissing` (legitimate fall-through) from parse errors (surface as a validation error so the user sees the broken file).

## Test coverage

- Config: shared `PickInitIdentity` and `ValidateAliasInfo` (spaces, unicode, missing prefix, empty fields).
- Daemon: path-free init, strict identity lookup, alias attach with metadata, alias conflict, reassign moves alias, reassign-without-alias rejection, local alias with spaces, rejects invalid kind / empty root_path.
- CLI: request shape via fake daemon (sends `project_identity` + alias info, no `start_path`), client-side conflict carries `code`, parse-error propagation, `kata health` honors `KATA_SERVER`.
- E2E: recording reverse-proxy in front of a real TCP daemon, fixing the wire contract end-to-end.

## State

Full test suite passes. Lint count unchanged from baseline (26 issues, all pre-existing).

Commits since `e250214`: `68b361b`, `aa3d69d`, `a6af41e`, `e5d8735`.

PR #24, opened 2026-05-06
docs: design and plan beads import

mariusvniekerk:
## Summary
- Add Beads import design spec
- Add approved implementation plan
- Establish upsert, import_mappings, daemon endpoint, and CLI behavior decisions

## Stack
- Base: main
- Next: #25 implementation

## Verification
- Docs-only PR

PR #25, opened 2026-05-06
feat: beads importer, first-class issue priority, schema cutover model

### LINKED ISSUE #16: Implement beads import command
Useful quality of life for beads users looking to exit

mariusvniekerk:
## Summary

This branch ships three coupled features plus a few quality-of-life fixes that fell out along the way.

closes #16 

### 1. Beads importer (`kata import --format beads`)

A normalized import pipeline: a generic daemon endpoint upserts external issues through `import_mappings`, with idempotent re-imports merged by source `updated_at`. The `kata import --format beads` command shells out to `bd export --no-memories`, fetches per-issue comments via `bd comments --json`, and submits a normalized batch.

- DB layer: `import_mappings` table (source, external_id, object_type → kata id) preserved through deletes, merges, and JSONL roundtrips.
- Daemon layer: `POST /api/v1/projects/{id}/imports` validates and atomically applies a batch; broadcasts the resulting events through SSE and hooks.
- CLI layer: `kata import --format beads` with auto-init prompt, idempotent re-imports, and a body footer that captures non-mapped beads metadata (id, type, original labels, timestamps, close reason).
- Beads' richer status vocabulary (open / in_progress / blocked / merged / etc.) collapses to kata's binary `open|closed`; non-trivial raw values are preserved as a `beads-status:<raw>` label so they stay queryable.
- Beads' integer priority maps directly to kata priority (see below); out-of-range values drop to nil.

### 2. First-class issue priority (0..4 nullable, 0 = highest)

- DB: nullable `priority INTEGER` column with `CHECK (priority IS NULL OR priority BETWEEN 0 AND 4)`. JSONL export/import preserves the field; older-schema cutovers leave it null.
- Daemon: dedicated `POST /actions/priority` endpoint, split events `issue.priority_set` and `issue.priority_cleared` carrying `priority` / `old_priority`, and digest/hook validation aware of the new types.
- Idempotency fingerprint folds in non-nil priority while preserving the legacy nil-priority canonical bytes — events stored before this branch keep matching. Priority is validated *before* the idempotency lookup so an out-of-range value is a 400 instead of being absorbed by a reuse path.
- CLI: `--priority 0..4` on `create`, `edit`, and `list` (plus `--max-priority` for range filters and `--priority -` to clear).
- TUI: new `prio` column in the list table and `priority:` row in the detail metadata. The `!` keystroke is bound in both views with slightly different UX:
  - **List view**: `!` arms a pending mode (status hint shows what's expected); the next keystroke commits — `0..4` sets, `-` clears, esc cancels. Single-stage so list editing stays terse.
  - **Detail view**: `!` opens a panel-local text-input prompt, mirroring the existing owner/parent/blocker/link prompts. Type `0..4` or `-` and press Enter.
  - Both bindings surface in the bottom hint bar (single line preserved) and in the `?` help overlay.

### 3. Schema cutover replaces the migration ladder

The repo previously had a `migrations/` directory implying a forward-only ALTER ladder, but kata's actual upgrade path is JSONL cutover (export old DB → fresh schema → atomic rename). The misleading subsystem is gone: there is now one `internal/db/schema.sql` that defines the current schema. `bootstrap()` applies it to fresh DBs in a single transaction; older DBs return `ErrSchemaCutoverRequired` so the JSONL path runs. `currentSchemaVersion` bumped to 6.

## Quality-of-life fixes

- Actor fallback chain now includes `$USER` between `KATA_AUTHOR` and `git config user.name`, so a fresh user gets a sane default (`wesm`) without having to set `KATA_AUTHOR`. Login names also read more cleanly than display names with spaces.
- A handful of golangci-lint findings on the import branch cleaned up (sha1 → sha256 for non-cryptographic label suffix, godoc on exported import types, test-helper context-first parameter order).
- Defensive nilaway guards on the import-batch state-map lookups: the first loop populates one entry per item so the lookups in subsequent loops always hit, but explicit `if state == nil { continue }` guards make the assumption visible at the call site.


mariusvniekerk:
This change is part of the following stack:

- #24
    - #25 ◀

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

wesm:
@mariusvniekerk ha I actually started having an agent look at this but didn't get very far. I was thinking it makes sense to add a "Priority" attribute to kata issues, thoughts?

wesm:
Note that this project is not doing database migrations (I actually want to tear out the migration subsystem altogether), we are using the JSONL export and import as the cutover to avoid having to do tedious migrations or backfilling

mariusvniekerk:
Yeah adding priority is good, also acts as a good tiebreaker for next task

wesm:
Let me work a little bit on this branch then, hold on

PR #28, opened 2026-05-08
Fix init with percent-encoded git remote URLs

njt:
## Summary

- `kata init --replace --project <name>` fails with "identity contains disallowed characters" when the git remote URL contains percent-encoded characters (e.g. `%20` for spaces in Azure DevOps org/project names)
- `NormalizeRemoteURL` now decodes percent-encoded characters and replaces spaces with hyphens, producing valid identities
- Affects both SCP-style (`user@host:path`) and HTTPS remote URLs

## Reproduction

```sh
# In a git repo with an Azure DevOps remote like:
# rtstnz@vs-ssh.visualstudio.com:v3/rtstnz/AI%20Experiments/AI%20Experiments

kata init --project my-project --name my-project --replace
# Error: alias.identity: identity contains disallowed characters:
#   "vs-ssh.visualstudio.com/v3/rtstnz/AI%20Experiments/AI%20Experiments"
```

## Test plan

- [x] Added test cases for SCP-style and HTTPS URLs with `%20` encoding
- [x] Verified existing `TestNormalizeRemoteURL` cases still pass
- [x] Full `go test ./...` passes (two pre-existing flaky timing tests in `internal/hooks` occasionally fail on both main and this branch)

### DIFF
--- a/internal/config/project_identity.go
+++ b/internal/config/project_identity.go
@@ -257,29 +257,39 @@ var scpLikeRE = regexp.MustCompile(`^([^@\s]+)@([^:\s]+):(.+)$`)
 
 // NormalizeRemoteURL strips credentials, normalizes SSH↔HTTPS, drops trailing
 // .git, and returns "host/path" form (e.g. "github.com/wesm/kata").
+// Percent-encoded characters are decoded and spaces replaced with hyphens so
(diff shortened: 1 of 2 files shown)


wesm:
LGTM

PR #29, opened 2026-05-08
cli: add --project selector flag

wesm:
## Summary

- Adds a global `--project <selector>` flag that targets a project by id, name, identity, or alias, bypassing workspace-based resolution.
- `events` and `digest` reject `--project` alongside `--project-id` / `--all-projects` so the chosen scope stays unambiguous.

## Why

Today the only way to scope CLI commands to a project is to `cd` into a workspace bound to it, or pass `--workspace <path>`. In remote-client mode that path may not exist on the client. `digest` and `events` already had `--project-id` (numeric only); other read/write commands had no escape hatch at all.

`--project` flows through the existing `resolveProjectSelector` helper used by `kata projects rename/merge/...`, so id/name/identity/alias matching works the same everywhere.

## Behavior

- `kata create --project <name> "title"` — works without cd, without `--workspace`.
- `kata list --project 7`, `kata show --project <name> 12`, `kata edit --project <name> 12 ...` — same shape.
- `kata events --project <name> --project-id 1` — fails fast with usage error.
- Workspace-bound flow is unchanged when `--project` is empty.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ needed. In a git workspace, the default project identity is derived from the
 remote URL. For a non-git workspace or an explicit shared identity:
 
 ```sh
-kata init --project github.com/example/product --name product
+kata init --identity github.com/example/product --name product
 ```
 
 Create and inspect issues:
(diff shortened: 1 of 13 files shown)


wesm:
Redoing this

PR #30, opened 2026-05-08
Consolidate relationship editing into create/edit flags

obra:
## Summary

Retires kata's eight dedicated relationship-editing commands (`link`, `unlink`, `parent`, `unparent`, `block`, `unblock`, `relate`, `unrelate`) in favor of matched flag pairs on `kata create` and `kata edit`:

| Add (create + edit) | Remove (edit only) |
|---|---|
| `--parent <ref>` (≤1, replaces existing) | `--remove-parent <ref>` (strict; must equal current) |
| `--blocks <ref>` (repeatable) | `--remove-blocks <ref>` (idempotent, repeatable) |
| `--blocked-by <ref>` (repeatable) | `--remove-blocked-by <ref>` (idempotent, repeatable) |
| `--related <ref>` (repeatable) | `--remove-related <ref>` (idempotent, repeatable) |

`<ref>` accepts `#N`, `N`, full ULID UID, or 8+ char UID prefix (case-insensitive).

A single `kata edit` PATCH is one DB transaction across title/body/owner/priority/links and emits up to three events: `issue.updated` (non-priority field changes), `issue.priority_set` / `issue.priority_cleared`, and a single aggregated `issue.links_changed` event for every applied add/remove.

The eight retired commands are deleted with no aliases (kata is in early preview).

## Design + spec

- `docs/superpowers/specs/2026-05-07-kata-relationship-flags.md` — full design.
- `docs/superpowers/specs/2026-04-29-kata-design.md` — banner + relationship sections superseded.

## Highlights of the implementation

- New `db.EditIssueAtomic` runs the whole PATCH in one TX; on any error nothing lands. Per-edge helpers (`addEdgeTx`, `removeEdgeTx`, parent set/replace) live alongside.
- Aggregated `issue.links_changed` payload mirrors the wire `changes` block plus parallel `*_uids` slices for stable peer identity across `kata reset-counter`.
- Single-peer aggregated events also populate envelope `related_issue_id` / `related_issue_uid` so consumers routing on envelope metadata stay informed; multi-peer events leave it NULL.
- Idempotent removes are idempotent end-to-end: missing target issue (numeric ref or UID-prefix) succeeds as a no-op without burning a daemon round-trip on a 400.
- Strict `--remove-parent` retains its loud-failure semantics: 409 `parent_mismatch` when the asserted ref doesn't equal current.
- Cycle detection on `set_parent` inside the same TX.
- Soft-delete-tolerant remove paths (link to a hidden peer is still removable).
- Purge / soft-delete preservation rule: when an issue is purged or soft-deleted, mutation events on OTHER live issues that reference it in payload are PRESERVED. Erasing that history is a worse outcome than leaving an orphan reference. Live-only JSONL exports scrub the envelope FK on aggregated events but keep the payload reference.
- TUI peer-match path is UID-aware so an event on a peer that shares a number with the watched detail (across counter resets) doesn't false-positive refresh.
- CLI ergonomics: `--blocks`/`--blocked-by` help text spells out "must finish before" semantics, `kata show` uses POV labels (`child: #N` / `blocked-by: #N`) instead of arrow direction, `kata create` echoes a `links: ...` summary like `kata edit` does, ULID prefixes accept any case, errors echo the offending ref (`link target #99 not found`).
- `kata list --json` now returns `blocks` / `blocked_by` / `related` peer numbers per row.

## Test plan

- [x] `go test ./...` green across `internal/jsonl`, `internal/db`, `internal/daemon`, `cmd/kata`, `internal/tui`, `internal/api`, `internal/hooks`, `e2e`.
- [x] Smoke test against a real binary in an isolated `KATA_HOME` exercising every flag, the retired-command "unknown command" path, atomic multi-flag PATCH, JSON wire-format expectations, ref forms, and JSONL export round-trip.
- [x] Three independent UX-feedback agents drove the new surface end-to-end (small refactor planning, bulk graph rearrangement, docs/help review). Findings folded back in.
- [x] 26 roborev review-fix-review iterations addressed 30+ findings ranging from cross-layer event propagation to wire-format consistency to error-message clarity.

## Known follow-ups

- Idempotency cross-upgrade asymmetry (Low): pre-kata#1 entries that hashed duplicate-bearing links can't reconcile with deduped retries within the 7-day window. Documented as a structural limitation in `tryIdempotencyMatch`. Self-heals after the window.
- Two pre-existing `internal/hooks` dispatcher tests fail on `main` (`TestDispatcher_Enqueue_RoutesToMatchingHooks`, `TestDispatcher_Shutdown_Timeout_ReportsInflight`); not introduced or modified by this PR.
- `make lint` is blocked by a tool/Go version skew predating this branch (golangci-lint built with Go 1.24, repo targets Go 1.26.1). `make test` is the verification path.


wesm:
This looks like a great simplification to me. Let me get #31 landed and then I'll rebase this and give it a tire-kicking

wesm:
thank you!

PR #33, opened 2026-05-11
Add --comment flag to mutation commands

obra:
## Summary

- `kata close 41 --comment 'fixed in abc123'` now appends the comment in one call instead of requiring a separate `kata comment 41 --body ...`. Agents (and humans) consistently forget the follow-up, so closes and edits land without context.
- The same `--comment TEXT` flag works on `close`, `reopen`, `edit`, `assign`, `unassign`, and `label add`/`rm`.
- Implementation is purely client-side: the mutation lands on its existing endpoint, then the CLI POSTs to `/comments`. No daemon changes, no new wire format. On comment failure the error names the issue so the user can retry with `kata comment <N> --body ...`.

## Why

The example use case is the exact prompt I keep seeing:

```
kata close 41 --comment "Fixed in e510319 — role=combobox/listbox/option, ..."
```

Before this PR that flag silently fails (`unknown flag: --comment`). After: it works on every mutation command where leaving a same-call note is useful.

## Test plan

- [x] `go test ./cmd/kata/ ./e2e/ -count=1` — all green
- [x] TDD: `TestClose_WithComment_AppendsComment`, `TestReopen_WithComment_AppendsComment`, `TestClose_EmptyComment_Rejected`, `TestAssign_WithComment_AppendsComment`, `TestUnassign_WithComment_AppendsComment`, `TestLabelAdd_WithComment_AppendsComment`, `TestLabelRm_WithComment_AppendsComment`, `TestEdit_WithComment_AppendsComment` — each watched fail before implementation
- [x] README and `kata quickstart` updated to document the flag

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,11 +194,17 @@ kata edit <issue-ref> [--title TEXT] [--body TEXT] [--owner NAME]
                   [--parent <ref>] [--blocks <ref>] [--blocked-by <ref>] [--related <ref>]
                   [--remove-parent <ref>] [--remove-blocks <ref>]
                   [--remove-blocked-by <ref>] [--remove-related <ref>]
+                  [--comment TEXT]
 kata comment <ref> [--body TEXT | --body-file PATH | --body-stdin]
-kata close <ref> --done --message <text> [--commit|--pr|--test|--reviewed|--evidence]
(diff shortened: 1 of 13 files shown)


wesm:
rebasing

mariusvniekerk (approved):

PR #36, opened 2026-05-11
Fix remote-client resolve regression (#35)

### LINKED ISSUE #35: Remote-client mode: `kata list` / `kata tui` fail with daemon-side stat error when `.kata.toml` is present
## Problem

In remote-client mode (`.kata.local.toml` `[server].url` or `KATA_SERVER` set), running `kata list`, `kata tui`, or any project-scoped command from a workspace with a committed `.kata.toml` fails:

```
$ kata tui
kata: validation: stat /srv/work/example: stat /srv/work/example: no such file or directory
```

The daemon is stat'ing the **client's** working directory on its own filesystem, where the path doesn't exist.

## Root cause

`POST /api/v1/projects/resolve` only accepts `name` or `start_path`. The CLI (`cmd/kata/create.go:resolveProjectIDAndName`) and TUI (`internal/tui/client.go:ResolveProject`) always send `start_path` unless `--project` was passed — even when `.kata.toml` is present locally with a valid name. The daemon then calls `config.DiscoverPaths` on that client-supplied path (`internal/daemon/handlers_projects.go:327`).

`kata init` already solved this: it derives `{name, alias}` client-side and POSTs path-free. `resolve` was left behind.

The existing behavior is pinned by `TestClient_ResolveProject_UsesStartPathForWorkspaceConfig` (`internal/tui/client_test.go:716`) and its CLI twin in `create_test.go:224`, both intentionally sending `start_path` so the daemon's alias-repair path can run. This breaks in remote mode.

## Proposed fix

Mirror `kata init`'s path-free flow on resolve:

1. Extend `ResolveProjectRequest` with an optional `Alias *AliasInput` field (same shape as `InitProjectRequest.Body.Alias`).
2. Daemon: when `alias` is supplied, look up by `alias.identity` (with rename/merge repair), fall back to `name`, attach alias on first-seen. No `DiscoverPaths` call.
3. Client (CLI + TUI, ideally via a shared `daemonclient` helper): when `.kata.toml` exists, send `{name, alias}` derived locally; when only `.git` exists, send `{alias}`; otherwise fall back to `start_path` (still works for local-daemon mode).
4. Move the .kata.toml rewrite-on-rename from daemon (`resolveByKataToml`) to client, since the daemon can no longer write the client's filesystem.
5. Invert the two pinning tests above; add daemon-side tests for alias-input repair and first-seen attach.

Affected files (read-only survey): `internal/api/types.go`, `internal/daemon/handlers_projects.go`, `internal/tui/client.go`, `cmd/kata/create.go`, plus tests.

## Workaround

Pass `--project <name>` on every invocation — the CLI sends `name` only and skips path resolution. No workaround for `kata tui`.

## Repro

1. Run a daemon on host A bound to a private IP.
2. On host B, in a directory `/srv/work/example` that doesn't exist on host A, create `.kata.toml` (`version = 1`, `[project] name = "X"`) and `.kata.local.toml` with `[server] url = "http://A:7777"`.
3. Run `kata list` or `kata tui`.

wesm:
## Summary

Restores path-free project resolution for project-scoped commands
(`kata list`, `kata tui`, `create`, `close`, …) in remote-client mode.

Issue #35: after 12ced3a collapsed `project_identity` into `project_name`,
the CLI and TUI fell through to `{start_path}` on every resolve when
`.kata.toml` was readable. A daemon on another host then ran
`DiscoverPaths` on a client path on its own filesystem and failed with
`validation: stat /path: no such file or directory`. Only `kata init`
was updated to the path-free wire shape; resolve was missed.

## Change

Mirror `kata init`'s path-free contract on resolve.

- **API** (`internal/api/types.go`): `ResolveProjectRequest.Body` gains
  optional `Alias *AliasInput`. Priority order: alias > name > start_path.
- **Daemon** (`internal/daemon/handlers_projects.go`): new
  `resolveByAliasInput` does alias-first lookup with a name fallback
  that attaches the alias on first-seen. Resolve never creates projects
  — an unknown alias without a matching name returns 404, so a git-only
  workspace whose alias isn't registered cannot resolve-by-convention
  from the git remote (init owns that path). Archived projects are
  rejected with 404 after the lookup so an alias pointing at a
  soft-deleted row doesn't slip through the active surface.
- **CLI + TUI**: derive wire shape locally. `--project` sends `{name}`.
  `.kata.toml` sends `{name, alias?}`. Git workspace without
  `.kata.toml` sends `{alias}`. Bare directory still falls back to
  `{start_path}` for local-daemon mode.
- **Client-side `.kata.toml` repair**: when the daemon returns a
  canonical name that differs from the local binding (project renamed
  daemon-side), the client rewrites `.kata.toml`. The daemon's own
  rewrite worked only in local mode.

The two tests that previously pinned `start_path` with a workspace
`.kata.toml` are inverted. New e2e regression `TestRemoteResolve_PathFreeWireShape`
asserts `kata list` hits `POST /api/v1/projects/resolve` with `{name, alias}`
and no `start_path` under `KATA_SERVER`.

Closes #35.

### DIFF
--- a/cmd/kata/create.go
+++ b/cmd/kata/create.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"os"
 	"strings"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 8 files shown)


wesm:
cc @cwensel, if you want to try this out and confirm it resolves the problem you were experiencing in #35

PR #44, opened 2026-05-15
Fix cutover wedge on pre-existing FK orphans

### LINKED ISSUE #43: Bug: Daemon auto-cutover wedges on pre-existing FK orphans, leaving DB unusable with no operator guidance
## What I was doing

Ran `kata list --json` against my main `~/.kata/kata.db` (schema v6) with a freshly-built `kata` binary (`gf30a60f`, expects schema v9).

## What happened

Daemon auto-start fails. The CLI prints:

```
kata: foreign_key_check: violations found
{"error":{"kind":"internal","message":"daemon failed to start within 5s","exit_code":1}}
```

After tracing the path:

1. `runDaemonWithListen` (`cmd/kata/daemon_cmd.go:201`) sees `schema_version < CurrentSchemaVersion()` and calls `jsonl.AutoCutover`.
2. `AutoCutover` does export → fresh DB → import. The importer runs `validateBeforeCommit` (`internal/jsonl/import.go:704`), which runs `PRAGMA foreign_key_check` and aborts the transaction on any violation.
3. My DB has 62 orphan rows (13 `events`, 39 `links`, 10 `comments`) referencing issue IDs (1–13, 19, 21, 60–65) that no longer exist in `issues`. They're residue from a corruption-recovery event on 2026-05-06/07 (see `kata.db.corrupt` + `kata.db-wal.corrupt` artifacts from that date in `~/.kata/`). Schema v6 didn't actively enforce these FKs at the moment those rows were left behind, so the broken refs sat undetected until the cutover importer ran the strict check.
4. The cutover aborts, the temp DB is removed, the source DB stays at v6, and every subsequent CLI invocation re-runs the cutover and re-fails. The daemon never starts. `kata` is unusable.

`PRAGMA integrity_check` on the source DB is `ok` — only referential integrity is broken. The orphan rows point at issues the app already can't see, so they're effectively dead weight.

## Why it matters

This is a one-way wedge for any operator who ran a corruption-recovery procedure on an older schema and is now picking up the latest binary. The only way out is hand-editing the SQLite file (`PRAGMA foreign_keys=OFF` + `DELETE FROM ... WHERE issue_id NOT IN (SELECT id FROM issues)`), which is not a thing most users will discover from the error message. The cutover *knows* exactly which rows are orphaned but discards that information and prints a single-line generic error.

Suggested directions (not prescriptive):

- Make the cutover importer's FK failure surface actionable detail — at minimum, the offending tables and rowids, and the count per child table — so the operator knows what they're looking at.
- Offer a recovery mode (e.g. `kata daemon repair` / `kata db fsck --fix-orphans`) that runs the FK check, prints the violations, and on confirmation moves the orphan rows into `lost_and_found` (already in the schema) before re-running the cutover. The current cutover's all-or-nothing semantics are correct for normal upgrades but leave no operator-facing path for "the source DB is already inconsistent."
- Consider running `PRAGMA foreign_key_check` on the *source* DB before starting the cutover, so the failure happens before the export/import dance and surfaces with clearer context ("source DB has 62 FK orphans, refusing to upgrade").

### Reproduction sketch

1. Build a kata DB at schema v6 (any cutover-eligible prior version works).
2. With `foreign_keys=OFF`, delete a handful of `issues` rows while leaving `events`/`links`/`comments` referencing them.
3. Upgrade to the current `kata` binary and run any command that auto-starts the daemon (`kata list`, `kata search`, etc.). Daemon fails to start, no recovery path is offered.

### Environment

- `kata gf30a60f`, built 2026-05-14, darwin/arm64
- Source DB: `~/.kata/kata.db`, schema_version=6, integrity_check=ok, 62 FK violations across `events`/`links`/`comments`

wesm:
Fixes #43.

## Summary

- `PreflightSourceFKs` opens the source DB read-only and runs `PRAGMA foreign_key_check` before cutover starts. The four known issue-child orphan classes (`events`, `comments`, `links`, `issue_labels`) are dropped or scrubbed at export; anything else halts cutover with `<table> rowid=R parent=P column=C` per row, so the operator sees exactly which rows are wedging the upgrade. Source DB and temp files are untouched on halt.
- When orphans were dropped, cutover prints one stderr line summarizing nonzero classes (e.g. `kata cutover: discarded 6 orphan rows from old DB (events: 1, comments: 2, links: 2, issue_labels: 1)`). Clean cutovers stay silent.
- The importer's `validateBeforeCommit` now reports each FK violation by row and resolved column instead of a bare count, so an unexpected schema in the wild produces an actionable error rather than `foreign_key_check: violations found`.
- Test coverage spans v3 source (legacy fixture) and v8 source (current schema with `meta.schema_version='8'`), the latter exercising the v5+ `import_mappings` halt path and the V8+ events scrub branch the v3 fixture cannot reach.


wesm:
cc @jesserobbins

wesm:
Appears hallucinated. Merging this @jesserobbins if you encounter more issues please reopen #43!

PR #47, opened 2026-05-20
Implement hub-and-spoke federation for shared kata projects

wesm:
## Summary

This PR implements kata federation end to end. Federation lets selected projects be shared across multiple kata daemons while keeping normal private projects fully local and unaffected. A hub daemon owns the canonical project stream, enrollment tokens, lease arbitration, purge/reset authority, and operational status. Spoke daemons keep local replicas, pull hub events, and can push local-origin work back to the hub when enrolled with the right capabilities.

Normal users with no federated projects should not be bothered by the feature: top-level help stays clean, the daemon does not read federation credentials, and the federation runner does not make network calls when no spoke bindings exist.

The implemented surface includes:

- replay-complete project events, deterministic fold/materialization, HLC ordering, and baseline snapshots
- pull replication from hub to spoke, including reset-required re-bootstrap after hub purge
- bidirectional spoke push with enrollment-token auth, origin binding, all-or-nothing ingest, idempotent retry, and quarantine for poisoned batches
- hub-authoritative federation leases with `kata federation lease acquire`, `kata federation lease release`, lease-gated issue mutations, pending/offline handling, force-release/steal admin operations, and lease display in `kata show`
- operational CLI/API support for `kata federation status`, enrollment audit/revoke, quarantine skip, purge/reset visibility, and lease violation surfacing
- stress and Docker Compose coverage for hub plus multiple spokes across real process/network boundaries

## How federation works

A project opts into federation by recording a hub-side binding and baseline snapshot. Each spoke receives a local replica binding and a credential file entry pointing at the hub. Enrollment tokens are stored hashed on the hub and are bound to a spoke instance UID, optional project scope, and capabilities such as `pull`, `push`, and `claim`/lease authority.

Spokes pull hub events after their pull cursor and apply them transactionally. Push-enabled spokes scan local-origin events after their push cursor and send batches to the hub. The hub verifies enrollment, origin, scope, content hashes, and live lease coverage before materializing accepted work. Same-hash retries are accepted so a lost response after commit does not wedge the spoke.

Leases are the synchronous arbitration point for federated writes. They are intentionally separate from ordinary ownership/assignment workflows: an owner is durable task responsibility, while a federation lease is short-lived, hub-authoritative write permission. A user or agent must hold the live lease for non-comment issue mutations on federated projects. Spokes refresh lease state from the hub when online and can use cached hard leases while offline; if work later arrives without current hub coverage, the hub records a `claim.violated` audit event rather than dropping data. The storage and audit schema still use the internal term `claim` for compatibility, but the user-facing CLI/API/docs call the concept a lease.

Daemon API tokens and federation enrollment tokens are separate. API tokens authenticate users/agents talking to a daemon, including local federation operator and lease commands when token identity is required. Enrollment tokens authenticate spoke-to-hub federation transport and are not general daemon API tokens.

## Schema and upgrade

This PR is a single schema bump from upstream schema `11` to schema `12`. Existing schema-11 databases upgrade through JSONL cutover. The v11 exporter stays on legacy comment/event projections for columns introduced here (`comments.uid`, event HLC fields, and event `content_hash`), and the importer backfills them into the schema-12 database.

## Consistency expectations

Federation is eventually convergent, not single-copy serializable. Once replicas have seen the same accepted event set, deterministic materialization makes their folded issue projections converge. Comments are preserved append-only, item-style changes such as labels and links converge by per-item add/remove semantics, and scalar fields converge by event/HLC order.

There are intentional consistency limits:

- spoke reads can be stale until the next pull
- local spoke writes are accepted locally before the hub serializes ingest
- offline cached hard leases can be superseded at the hub and later produce `claim.violated`
- `claim.violated` is best-effort at ingest time, not historical proof of authorization at edit time
- poisoned push batches require operator quarantine/skip decisions
- hub purge creates reset boundaries, and spokes may need to re-bootstrap
- enrollment tokens authorize spoke daemons, not a full multi-tenant user/role model

Use a shared daemon instead of federation when a team needs immediate single-copy reads, centrally enforced user authorization, synchronous conflict prevention, strict online lease enforcement, or destructive administrative actions to become visible to every user immediately. Federation is for mixing private local projects with selected shared/federated projects while preserving local-first availability.

## Docs

The implemented behavior and limitations are documented in `docs/federation.md`. The original design context remains in `docs/superpowers/specs/2026-05-20-kata-federation-design.md`.

## CI and verification

CI runs the full PR gate as separate jobs:

- `go test -p 1 ./...`
- `golangci-lint` using `.golangci.yml` (including `govet`)
- `make nilaway`
- `make test-stress`
- `make test-federation-docker`

Local verification on this branch passed `make test`, `make vet`, `make lint`, `make nilaway`, `make test-stress`, and `make test-federation-docker`.


blaine:
I would consider using atproto for this. Identity, a well-defined position on evolutionary schemas, a data flow that works the same as the hub/spoke model you've described but with robust positions on how to implement the subtle kappa architecture pieces.

Easy to set up a local-only network (private PLC + PDS, fast local copy is roughly a local appview), robust story for extending the work into the open (just use the public PLC), private data on atproto story "coming very soon"

erans:
How would the "claim" work? it means the local instance needs to be in constant communication with the hub.
Or, during a request for claim they MUST sync with the hub and claim it.

blaine:
> How would the "claim" work? it means the local instance needs to be in constant communication with the hub. Or, during a request for claim they MUST sync with the hub and claim it.

Yup; with atproto, there's a relay with a cursor that effectively creates a shared vector clock. I'm not aware of synchronized systems being built on top of atproto, and while individual PDSes are merkelized, the whole network is not guaranteed to be consistent and/or convergent. That said, there are a few ways to skin this particular cat:

- The hub is just one PDS. A central agent handles writes to the central PDS, this is already guaranteed ordered (like a git repo).
- Each agent has a PDS, and claims are not guaranteed exclusive (may result in duplicated work).
- Each agent has a PDS, but a central agent handles locks.
- Each agent has a PDS, and all agents use a shared relay with a consistent cursor. Locks on claims are obtained by writing optimistic lock requests to their own PDS and waiting for the lock to come back on the relay. If there were no requests for the claim in question that precede the awaited optimistic lock on the [ordered] relay, then the agent can go ahead, otherwise try again.
- I haven't done it, but I think the previous approach could be enhanced to be deadlock-free by implementing the semantics here: https://notes.brooklynzelenka.com/blog/surelock/  @expede, what say you?

cpcloud:
_AI disclosure: the analysis and this comment were drafted by Claude, reviewed by me before posting._

Compared this against atproto: don't adopt it. The design holds up, and two atproto-adjacent ideas would land as later additions.

The shapes look similar on the surface. The SSE firehose lines up with subscribeRepos, UID + origin_instance_uid with at://did/nsid/rkey, the replay horizon with a repo commit. The substance diverges in five places:

- **Identity unit.** atproto's replication unit is the per-DID repo; kata's fold input is the project event set, written by many actors. A single project-DID makes every record sign with the same key, so signatures stop binding to the actual author. One-repo-per-(user, project) splits the project's events across N repos, so no replica natively holds the set §6.4 folds over. You'd build the stitching layer that atproto doesn't provide.
- **Cross-repo ordering.** atproto orders within a repo by parent-CID; across repos, by relay receive-time. Receive-time inverts under network reordering: if A writes after seeing B's event but the relay sees A's first, the order flips. The HLC in §6.3 advances on apply (local clock = max(local, foreign) + 1), so A's clock is forced past B's and the causal order survives whatever the network does.
- **Schema evolution.** Lexicons require strict validators at every relay and a coordinated upgrade across all of them when fields change. The kata cutover transforms a JSONL dump in one writer pass on the upgrading daemon; no N-party rollout.
- **CRDT semantics.** atproto's merge unit is the whole record; concurrent edits to different fields clobber each other unless the client read-modify-writes the full record under a lock. Per-field LWW, per-element tombstones, and structural-absence metadata deletion (§6.2, §6.6) live in the fold engine either way, so atproto's record store doesn't reduce that work.
- **Synchronous arbitration.** atproto is push-to-relay with no atomic claim mechanism. §9.2's ≤1-holder guarantee needs a synchronous endpoint at the hub backed by SQLite's single-writer transaction; atproto has no equivalent serialization point.

Two atproto-adjacent ideas worth doing as later additions, not as foundation:

1. **Per-instance event signing.** D8's bearer + origin-binding is enough on a private network, but a leaked token lets anyone forge events from the bound origin. Each daemon already has an `instance_uid`; give it a keypair, sign each locally-originated event's hash, and the hub can verify provenance independently of the transport. Fits Phase 2 alongside §7.4 without touching fold.
2. **Hash-chained event log.** Each event commits to the previous event's hash. A spoke verifies "I have exactly events 0..N" with one hash compare. MST exists to prove inclusion of arbitrary records (random-access proofs), which kata doesn't need; a single chain handles sequential catch-up.

One piece worth filing away: DID-based actor identity. If D9's free-form actors ever need to outlive a host, DIDs ship with a resolution and key-rotation ecosystem rather than requiring an in-house one. Adding it as a side field on actor records touches only the actor column; retrofitting kata onto atproto touches storage, transport, identity, and schema at once.

wesm:
Regarding atproto, from my discussion with codex about this to clarify my thinking:

"Kata is not trying to be a trustless federated social network. It is
trying to keep a shared project reasonably consistent across
cooperating instances. That means we can choose a much simpler
consistency model: project events are replicated, folds are
deterministic, and operations that require exclusivity, like claims,
go through a small synchronous serialization point.

A claim is not a cryptographic proof of universal network state. It is
a practical lease/lock backed by the hub’s single-writer transaction
boundary. That does mean an instance must talk to the hub when it
wants exclusive work, but that is acceptable because claim acquisition
is exactly the kind of operation where synchronous arbitration is
useful. It does not imply every read, fold, or local action requires
constant communication.

atproto solves a different problem: independently operated personal
repositories, public identity, replayable social data, and relay
fanout. It does not give cross-repo serializability or atomic claim
arbitration. If we built on it, we would still need our own fold
semantics, conflict handling, project-level ordering, and lock
service. At that point we would have adopted a large substrate while
still building the parts kata actually needs.

Some atproto-adjacent ideas are worth borrowing later, especially
per-instance event signing, hash-chained event logs, or DID-shaped
actor identity. But those are incremental hardening features. They are
not a reason to make atproto the foundation."

wesm:
Codex follow-up on the federation review feedback:

I updated the spec to keep the trusted-network/default-agent path practical while still closing the real integrity gaps.

Plain-language summary of the changes:

- The federation auth boundary is now stated explicitly as spoke daemon -> hub, not individual agent -> hub. In the normal setup, ephemeral agents are thin clients of a long-lived spoke daemon. The spoke enrolls with the hub once, so agents do not need per-agent hub tokens.
- Token scope is now proportional to the deployment. A trusted single-team hub can use a coarse enrollment grant, such as all federated projects on that hub. Mixed or untrusted hubs can narrow grants by project/capability. This avoids forcing full multi-tenant auth machinery into the trusted-network case.
- Ingest now has content-level project validation. Even with a coarse grant, every incoming event must resolve its project/issue/link/move/metadata/recurrence references inside the authorized project set before the hub inserts or folds it. This protects the event log from malformed or cross-project payloads.
- Claim bypass no longer depends on the free-form `--as` actor string. Human/admin bypasses must come from trusted spoke policy or a server-authenticated role/capability/client kind. If a deployment has no such signal, federated non-comment work mutations require claims.
- Metadata deletes are now prefix-aware subtree tombstones. A later clear of `/foo` hides older descendant writes like `/foo/bar`; a newer descendant write can recreate the subtree. This matches the existing nearest-common-pointer LWW rule.

The intent is to keep federation aimed at trusted users and agents on private networks, while making the spec precise enough that implementation does not accidentally create cross-project injection, claim impersonation, or metadata merge holes.

wesm:
Human Wes here: thanks all the comments. I updated the PR description with notes about atproto and I'll pause here and wait for more feedback before moving forward. A mantra here is to protect and preserve simple/lightweight kata user experience (the average user can pretend that this feature does not exist), and so the idea is convenient shared projects with a hub-and-spoke model, trading strong consistency guarantees for keeping things simple.

wesm:
I created a preliminary implementation of the federation spec. For clarity, you would not use this as an alternative to an ACID-compliant database to implement durable and exactly-once execution for example, but it should provide a simple and serviceable foundation for humans and agents to collaborate on a shared kata project in a team setting while retaining their private projects locally. Human and clanker feedback on the implementation details welcome.

PR #63, opened 2026-05-28
Bind 0.0.0.0:$PORT for PaaS hosting (Heroku, Cloud Run, Fly.io)

cpcloud:
## Summary
- Daemon binds `0.0.0.0:$PORT` when `PORT` is in the environment and neither `--listen` nor config `listen` is set — the contract Heroku originated and that Cloud Run, Render, Fly.io, Railway, and App Engine all follow.
- An auto-start marker (`KATA_AUTOSTART=1`, stamped by `daemonclient.autoStart` on the spawned child) keeps a stray `PORT` in a developer's shell from hijacking implicit local daemons.
- `requireNonPublic` now permits the unspecified wildcard (`0.0.0.0` / `::`); exposure stays gated by `checkAuthStartup`, which still refuses non-loopback binds without a token plus `trust_private_network` (or `--insecure-readonly`).
- `AGENTS.md` gains a "Hosted mode" section covering the env contract, health probes, shutdown, and persistence / single-instance caveats.
- New e2e coverage exercises the real serve path, SIGTERM shutdown, and the autostart suppression.

## Test plan
- [x] `go test ./...` green
- [x] `golangci-lint run` clean on touched packages
- [x] `e2e.TestPortEnvBind_ServesAndShutsDownCleanly` — real bind on `0.0.0.0:$PORT`, `/api/v1/health` returns 200 unauthenticated, `/api/v1/projects` returns 401 without token and not-401 with one, daemon exits cleanly on SIGTERM
- [x] `e2e.TestPortEnvBind_AutostartMarkerSuppresses` — `KATA_AUTOSTART=1` + `PORT` keeps the PORT address unbound and the daemon falls back to the Unix socket path

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -96,3 +96,29 @@ A daemon can serve clients on other hosts over a private network:
   `kata init` falls back to a path-based request only when none of
   those sources are available, so the daemon (or its absence) emits
   the existing validation error.
+
+## Hosted mode (`$PORT` convention)
+
+When no `--listen` flag and no config `listen` are set, the daemon binds
+`0.0.0.0:$PORT` if `PORT` is in the environment. This is the contract
+Heroku originated and that Cloud Run, Render, Fly.io, Railway, App
+Engine, and other PaaS runtimes all follow: the platform picks a port,
+injects it via `PORT`, and expects the process to bind every interface
+at `0.0.0.0:$PORT`. The auto-start child sets `KATA_AUTOSTART=1` so a
+stray `PORT` on a developer's machine does not flip implicit local
+daemons onto wildcard TCP.
+
+- Required env on the host: `KATA_AUTH_TOKEN=<token>` plus
+  `KATA_TRUST_PRIVATE_NETWORK=1`. Without both, the daemon refuses the
+  non-loopback bind — the platform terminates TLS upstream, so the
+  operator must explicitly assert trust in the container's network path.
+- `KATA_HOME` must point at a writable path (e.g. `/tmp/kata`); the
+  daemon writes runtime files and the SQLite DB under it. Local container
+  disk is ephemeral, so data does not survive instance recycling unless
+  `KATA_HOME` is backed by a mounted volume or shared storage.
+- Health: `GET /api/v1/health` and `GET /api/v1/ping` are unauthenticated
+  and suitable for platform liveness / readiness probes.
+- Shutdown: the daemon handles SIGTERM gracefully (up to 10s for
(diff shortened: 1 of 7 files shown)


cpcloud:
Both High findings in the latest roborev-ci review are false positives — flagging here so a future reviewer doesn't have to re-derive.

**Finding 1 — "`strings`/`strconv` not added to imports, `cmd/kata` won't compile"**

Both are already imported at `cmd/kata/daemon_cmd.go:11-12` (they were in use elsewhere in the file before this change). The passing `Go tests` CI job is the proof — a missing import is a hard compile error.

**Finding 2 — "no code reads `KATA_TRUST_PRIVATE_NETWORK`, so PaaS mode can't start"**

Both halves are wrong. The env var is read at `internal/config/daemon_config.go:119-120`:

```go
if EnvTruthy("KATA_TRUST_PRIVATE_NETWORK") {
    cfg.Auth.TrustPrivateNetwork = true
}
```

And the auth-startup gate permits the hosted bind when token + trust are both set, at `internal/daemon/auth.go:99`:

```go
if p.Token != "" {
    if p.TrustPrivateNetwork {
        return nil
    }
    ...
}
```

The new e2e test `TestPortEnvBind_ServesAndShutsDownCleanly` exercises this path end-to-end: it sets `KATA_AUTH_TOKEN=t` + `KATA_TRUST_PRIVATE_NETWORK=1` + `PORT=<port>`, then asserts the daemon binds `0.0.0.0:$PORT`, serves `/api/v1/health` 200 unauthenticated, gates `/api/v1/projects` (401 without token / 200 with), and exits cleanly on SIGTERM. If the auth gate were rejecting the bind, the test would fail at startup.

AGENTS.md
```diff
@@ -96,3 +96,29 @@ A daemon can serve clients on other hosts over a private network:
   `kata init` falls back to a path-based request only when none of
   those sources are available, so the daemon (or its absence) emits
   the existing validation error.
+
+## Hosted mode (`$PORT` convention)
```
mariusvniekerk, line 100: 25 lines of AGENTS.md is _very_ expensive for this relatively limited set of context stuff,  most of this belongs in a different doc that the agent can retrieve via progressive disclosure.

cpcloud: Maybe our top-level AGENTS.md needs some more admonition about this kind of thing so we don't have to think too hard about keeping this particular thing on the rails.

cpcloud: Addressed in #64

PR #65, opened 2026-05-28
Add simple token identity for shared-daemon attribution

wesm:
## Summary

Adds a DB-backed token identity layer so a shared kata daemon can attribute mutations to a server-verified actor instead of trusting the client-supplied `actor` field. This is a minimal first step toward shared-server mode — one `api_tokens` table with the actor on the token row, no users table, roles, or impersonation.

Behavior is gated behind a new `[auth].require_token_identity` flag (default `false`). With it off, local no-auth and single-token modes are unchanged and no DB token lookup happens on writes.

## How token credentials work mechanically

`kata tokens create --actor <actor> --name <label>` is a provisioning step. During that request, the daemon generates a high-entropy plaintext bearer token, returns it exactly once in the create response, and stores only `sha256(token)` in the `api_tokens` projection. The plaintext token is not stored in SQLite, events, logs, JSONL export, or any later list/resolve response.

After creation, the operator/user must save that one-time plaintext token on the client side, for example in the client's kata config or an environment variable:

```toml
[auth]
token = "kata_..."
```

```sh
export KATA_AUTH_TOKEN=kata_...
```

When a CLI command talks to a remote daemon, the CLI loads the token from that local client-side config/env source and sends it as:

```http
Authorization: Bearer kata_...
```

The daemon hashes the presented bearer value with SHA-256 and looks up `api_tokens.token_hash`. If it finds a non-revoked row, that row supplies the canonical actor for the request. In identity mode, that server-derived actor overrides any actor string sent in the request body.

There is no "retrieve my existing token" API. If the plaintext token is lost, the recovery path is to revoke that token row and create a replacement. This is intentionally similar to a GitHub personal access token flow, not an OAuth login/session flow and not SSH public-key request signing.

## What's included

- **Token admin CLI**: `kata tokens create --actor <a> --name <n>`, `kata tokens list`, `kata tokens revoke <id>`. The plaintext bearer token is intentionally printed exactly once at creation so an operator/user can copy it into local client config or an environment variable; only its SHA-256 hash is stored, and lost tokens must be revoked and recreated.
- **Token admin API**: `POST /api/v1/tokens`, `GET /api/v1/tokens`, `POST /api/v1/tokens/{id}/actions/revoke`, restricted to the bootstrap token or loopback/no-auth access (DB tokens cannot mint or revoke tokens).
- **Identity middleware** (when enabled): the bearer is matched constant-time against the bootstrap token, otherwise looked up by hash in `api_tokens`. GETs and SSE accept either token; mutations require a DB token and the daemon derives the actor from it, ignoring any actor in the request body. The bootstrap token can read and manage tokens but cannot perform attributed writes. A client-declared `source=tui` no longer grants the close-policy bypass for remote identity-mode requests.
- **Event-sourced lifecycle**: `token.created` / `token.revoked` events are recorded against a hidden system project so tokens survive JSONL export/import and schema cutover. These events are excluded from ordinary project lists, stats, cross-project polls, SSE, and audit surfaces. Plaintext never appears in events, logs, or the DB.
- **Schema**: new `api_tokens` table (a projection rebuilt from token events), schema-version bump with completeness guard, and lazy `last_used_at` updates (skipped if updated within the last hour).
- **Error codes**: `auth_required` (401), `token_invalid` (403), `bootstrap_token_write_forbidden`, `token_admin_forbidden`.
- `--insecure-readonly` and `require_token_identity = true` are rejected together at startup.

Design and implementation notes are in `docs/superpowers/specs/2026-05-28-simple-token-identity-design.md`.


internal/daemon/identity.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	return fallback
+}
+
+func tuiBypassAllowed(ctx context.Context, source, reason, requestActor string) bool {
+	if source != "tui" || reason != "done" {
+		return false
+	}
+	if p, ok := PrincipalFromContext(ctx); ok && p.Kind == PrincipalDBToken {
+		return strings.TrimSpace(requestActor) == p.Actor
```
cpcloud: **P1 — this `source=tui` bypass contradicts the spec and the plan.**

The design doc states the bypass must not be honored merely because the JSON body claims a TUI origin, and that eligibility must derive from trusted request context, not from `in.Body.Source`; until that context exists, remote identity-mode requests get no `source=tui` evidence bypass. The implementation plan shows the `PrincipalDBToken` branch returning `false`.

Here the bypass is granted whenever `in.Body.Actor` equals the token actor — but `actorFor` already forces attribution to `p.Actor`, so any client trivially satisfies this by echoing its own actor. The check adds no protection against the threat the spec describes. Effect: a remote DB-token client can close any reachable issue with a body of `source=tui` plus its own actor and skip `ValidateCloseInput` and `validateEvidenceTargets` (no message, no typed evidence). The structural guards (parent-close, throttle, repeated-message) still apply, so this is audit-integrity erosion rather than privilege escalation.

Fix: make the `PrincipalDBToken` branch return `false` (matching the plan), drop the `requestActor` parameter, and flip the test that pins this. If the actor-match is genuinely the intended policy, update the design doc instead — right now code and spec disagree.

internal/daemon/close_guards_test.go
```diff
@@ -22,6 +24,54 @@ type throttledEventRecord struct {
...
+		map[string]any{"actor": "someone_else", "source": "tui"},
+		map[string]string{"Authorization": "Bearer alice-token"})
+	assertAPIError(t, resp.StatusCode, bs, http.StatusBadRequest, "validation")
+
+	got, err := env.DB.IssueByID(context.Background(), issue.ID)
+	require.NoError(t, err)
+	assert.Equal(t, "open", got.Status)
+}
+
+func TestCloseIssue_IdentityModeAllowsTUIBypassForMatchingTokenActor(t *testing.T) {
```
cpcloud: This test pins behavior that contradicts the design spec — remote identity-mode requests should not receive the `source=tui` evidence bypass (see the note on `identity.go`). It asserts a DB-token client closes with no message or evidence by echoing its own actor. If the bypass is corrected to return `false` for DB tokens, this should assert a `400 validation` error instead.

internal/db/tokens.go
```diff
@@ -0,0 +1,324 @@
...
+// ResolveAPIToken resolves a plaintext bearer token to its active token row.
+func (d *DB) ResolveAPIToken(ctx context.Context, plaintext string) (APIToken, error) {
+	hash := tokenHash(plaintext)
+	tok, err := scanAPIToken(d.QueryRowContext(ctx,
+		apiTokenSelect+` WHERE token_hash = ? AND revoked_at IS NULL`,
+		hash))
+	if err != nil {
+		return APIToken{}, err
+	}
+	res, err := d.ExecContext(ctx, `
```
cpcloud, line 251: **P2 — this advisory `last_used_at` write can reject a valid token.**

Any error from this UPDATE (or the `RowsAffected` call below) is returned to the caller, and `requireIdentityBearer` maps every `ResolveAPIToken` error to a `403 token_invalid`. So a transient `SQLITE_BUSY` on this metadata write — which, unlike the mutation handlers, is not wrapped in `RetryLockContention` — rejects an otherwise-valid credential with a misleading diagnostic. The field is advisory and already throttled to about once per hour.

Fix: make the bump best-effort — on UPDATE failure return the already-resolved `tok` with a nil error, and re-scan only when `RowsAffected > 0`.

internal/daemon/auth.go
```diff
@@ -71,11 +83,49 @@ func requireBearer(p authPolicy) func(http.Handler) http.Handler {
...
+		return
+	}
+	if tokenStore == nil {
+		api.WriteEnvelope(w, http.StatusInternalServerError, "internal",
+			"token identity requires a database")
+		return
+	}
+	tok, err := tokenStore.ResolveAPIToken(r.Context(), presented)
+	if err != nil {
+		api.WriteEnvelope(w, http.StatusForbidden, "token_invalid", "token invalid")
```
cpcloud, line 120: **P2 — every `ResolveAPIToken` error collapses to `403 token_invalid`.**

`ResolveAPIToken` returns `db.ErrNotFound` for unknown or revoked tokens (correctly a 403) but also wraps real query/scan/lock faults, including the `last_used_at` write flagged in `tokens.go`. A genuine DB problem then presents to the client, and in the logs, as a bad token — sending operators to debug the wrong thing.

Fix: branch on `errors.Is(err, db.ErrNotFound)` for the 403; map other errors to 500/503 with the underlying cause logged, mirroring the split already used in `handlers_tokens.go`.

internal/daemon/auth.go
```diff
@@ -92,6 +142,9 @@ func requireBearer(p authPolicy) func(http.Handler) http.Handler {
 // daemon with a TLS-terminating reverse proxy and bind the daemon to a
 // Unix socket or 127.0.0.1.
 func checkAuthStartup(listen string, p authPolicy) error {
+	if p.RequireTokenIdentity && p.InsecureReadonly {
```
cpcloud, line 149: **P2 — no startup guard for identity mode without a bootstrap token.**

`checkAuthStartup` refuses `require_token_identity` together with `--insecure-readonly`, but not `require_token_identity` with an empty `Token`. In that configuration every request enters `requireIdentityBearer`, the bootstrap branch is gated on a non-empty `p.Token`, and an empty token DB returns `403 token_invalid` for everything — including `POST /api/v1/tokens`, so the operator cannot even seed the first user token. The daemon starts without complaint: a silent lockout.

Fix: add a guard here that returns an error when `RequireTokenIdentity` is set but `Token` is empty, so the daemon refuses to start in that configuration.

internal/daemon/handlers_ownership.go
```diff
@@ -18,7 +18,10 @@ func registerOwnershipHandlers(humaAPI huma.API, cfg ServerConfig) {
 		Method:      "POST",
 		Path:        "/api/v1/projects/{project_id}/issues/{ref}/actions/assign",
 	}, func(ctx context.Context, in *api.AssignRequest) (*api.MutationResponse, error) {
-		actor := strings.TrimSpace(in.Body.Actor)
+		actor := strings.TrimSpace(actorFor(ctx, in.Body.Actor))
```
cpcloud: **P3 — these three ownership handlers diverge from the shared `attributedActor` helper.**

assign/unassign/claim inline `strings.TrimSpace(actorFor(...))` plus `ensureAttributedWriteAllowed` plus `validateActor`, while the other ~22 mutation handlers call `attributedActor`, which does not trim. Since `validateActor` only checks non-empty, in no-auth/static mode an actor such as `alice ` (trailing space) is persisted trimmed here but raw elsewhere — inconsistent attribution — and a future change to `attributedActor` will silently miss these three. Collapse to `actor, err := attributedActor(ctx, in.Body.Actor)`; if trimming is desired, push it into the shared helper.

internal/db/tokens.go
```diff
@@ -0,0 +1,324 @@
...
+}
+
+// EnsureSystemProject creates the hidden project used to anchor daemon-global
+// token lifecycle events. It is idempotent so every Open can call it after the
+// normal schema bootstrap path.
+func (d *DB) EnsureSystemProject(ctx context.Context) error {
+	_, err := d.ExecContext(ctx, `
+		INSERT INTO projects(uid, name)
+		VALUES(?, ?)
+		ON CONFLICT(name) DO UPDATE SET uid = excluded.uid
```
cpcloud: **P3 — `ON CONFLICT(name) DO UPDATE SET uid` can silently rewrite an existing project UID.**

This runs on every `Open`. If a `.kata-system` row ever exists with a different UID (a migrated or hand-edited DB, or a future code path), its UID is silently overwritten and any events keyed to the old UID are orphaned. Latent today because `CreateProject` rejects the reserved name. Prefer `ON CONFLICT(name) DO NOTHING` so a mismatch fails safe instead of being masked.

internal/jsonl/import.go
```diff
@@ -107,6 +116,109 @@ func readMetaInstanceUID(ctx context.Context, tx *sql.Tx) (string, error) {
...
+				createdAt, rec.TokenID)
+			if err != nil {
+				return fmt.Errorf("replay token.revoked %d: %w", rec.TokenID, err)
+			}
+			n, err := res.RowsAffected()
+			if err != nil {
+				return fmt.Errorf("replay token.revoked rows affected: %w", err)
+			}
+			if n == 0 {
+				return fmt.Errorf("replay token.revoked %d: token not found", rec.TokenID)
```
cpcloud, line 216: **P3 — the malformed-event branches in token replay are untested.**

This token-not-found path, and the missing-field branches above it, abort the entire import, but the roundtrip and cutover tests only feed well-formed events. Add a test that imports a hand-built JSONL stream containing a `token.created` missing its `token_hash` and a `token.revoked` referencing an unknown `token_id`, asserting `Import` fails with these errors.

internal/db/schema_completeness_test.go
```diff
@@ -19,7 +19,7 @@ func TestAllSchemaTablesExist(t *testing.T) {
 	d := openTestDB(t)
 	wanted := []string{
 		"projects", "project_aliases", "issues", "comments",
-		"links", "issue_labels", "events", "purge_log",
+		"links", "issue_labels", "events", "api_tokens", "purge_log",
```
cpcloud, line 22: **P3 — this completeness guard is one-directional.**

`TestAllSchemaTablesExist` asserts that each name in the hardcoded list exists, but nothing asserts the inverse — that every table in `sqlite_master` is on the list — so a future table added without updating this list passes silently. `api_tokens` is correctly listed here, so there is no bug in this PR; worth tightening with a `SELECT name FROM sqlite_master WHERE type='table'` diff against the expected set.

cpcloud (commented):
## Review: DB-backed token identity

Build and the changed-package tests pass (unshuffled) and `golangci-lint` is clean. The core invariant holds uniformly: every mutation handler resolves the actor server-side via `attributedActor`/`actorFor` and ignores the request-body actor. Token generation (crypto/rand, 256-bit, SHA-256, `kata_` prefix), constant-time bootstrap compare, revoked/unknown/missing handling, transaction atomicity, plaintext non-leakage, and the `--insecure-readonly` rejection all check out.

Inline comments carry the specific findings. The merge-blockers:

- **P1** — the `source=tui` close bypass is granted to remote DB-token clients (`internal/daemon/identity.go`), contradicting the design spec and the implementation plan.
- **P2** — the token resolve path is fail-closed and error-collapsing (`internal/db/tokens.go`, `internal/daemon/auth.go`): a transient DB error rejects a valid token.
- **P2** — identity mode with no bootstrap token silently locks the operator out (`internal/daemon/auth.go`).

### General findings (not tied to one line)

**P2 — the actor-override invariant is under-tested.** It is the security property this PR exists to enforce, yet identity-mode tests cover only issue-create, close, and reopen. The other ~23 mutation endpoints (comments, labels, links, ownership, priority, metadata, delete/restore/purge, recurrences, move, imports, project remove/restore/alias-detach, issue-edit) run the same helpers with no test asserting the body actor is ignored and the bootstrap token is refused. Add a table-driven test across every endpoint; the destructive and ownership paths matter most.

**P3 — system-project hiding is unverified on the audit, digest, and stats HTTP surfaces.** They inherit the `EventsInWindow`/`BatchProjectStats` filters, but no test asserts `token.*` events or the system project stay out of those responses (there is no audit-handler test at all). A future direct query would regress this silently.

**P3 — `make test` is non-deterministic.** The Makefile runs `-shuffle=on`; under shuffle, `cmd/kata` tests fail (e.g. `TestSmoke_ExportImport` with `conflicting output modes`, and `TestDaemonStart_ListenFlagRejectsMalformed`) because some tests do not reset the global output-mode flag. Pre-existing pattern — this PR's token tests reset flags correctly via `requireCmdOutput` — but it makes the default test command flaky and is worth a separate cleanup.

### Minor / informational (kept out of the diff)

- `actorFor` silently substitutes the token actor for a mismatched body actor — matches the documented design; a 400 on mismatch would be friendlier but changes the contract.
- `token.created` payloads carry the SHA-256 hash; safe given system-project hiding and required for JSONL replay.
- JSONL import preserves each item's source author (historical authorship) while event actors are token-derived — correct split.
- `rollbackUnlessCommitted` (`internal/db/tokens.go`) just calls `Rollback` and relies on `ErrTxDone`; the name overstates it and diverges from the package's inline-defer idiom.
- The `bootstrap_token_write_forbidden` message is duplicated in `auth.go` and `identity.go`; extract a const so they cannot drift.

wesm:
Addressed in 61cda7d.

- DB-token principals can no longer use `source=tui` to bypass close-policy validation; that bypass is now local/trusted only.
- Bootstrap identity is now allowed through middleware for read-like POSTs/token admin, while attributed writes are still rejected at the handler attribution boundary.
- Token lookup now distinguishes `ErrNotFound` from real DB failures, and `last_used_at` updates are best-effort.
- `require_token_identity` now requires a configured bootstrap token at startup.
- Ownership handlers now use the shared attribution helper.
- System project remove/restore is blocked, and `EnsureSystemProject` no longer rewrites a conflicting sentinel UID.
- Added malformed token-event import tests and tightened schema table completeness coverage.
- Also fixed the shuffled `cmd/kata` test leak by resetting global CLI flag state in `newRootCmd`.

Verified with `make test`, `make lint`, full `go test ./... -count=1`, the failing shuffle seed for `cmd/kata`, and pre-push nilaway.

cpcloud (approved):
Verified the latest commit (`1cec869`) against my earlier review — all findings are addressed:

- **P1** `source=tui` close bypass now returns `false` for DB-token principals (`identity.go`), and the close-guard test is flipped to assert `400` for the echo-own-actor case.
- **P2** `last_used_at` writes are best-effort (a failed update no longer rejects a valid token); token-resolve errors split `ErrNotFound`→403 from real DB faults→500, with `TestAuthMiddleware_IdentityModeTokenLookupError500` covering it.
- **P2** startup now refuses identity mode without a bootstrap token.
- **P3** ownership handlers use the shared `attributedActor` helper; `EnsureSystemProject` uses `ON CONFLICT DO NOTHING` + a fail-loud UID check; malformed token-event import and bidirectional schema completeness are now tested.
- Actor-override is covered across create/assign/close and table-driven project mutations, and the system project is blocked from remove/restore/merge. Bootstrap read-like POSTs (`/projects/resolve`) work.

`go build`, the full test suite, and `golangci-lint` are green locally.

One non-blocking follow-up: `make test` (with `-shuffle=on`) still has a residual flake in `TestDaemonStatus_NoDaemonReportsAbsent` — it runs `newDaemonCmd()` directly without `resetFlags(t)` like its siblings, so it occasionally reads leaked output-mode state. Pre-existing and substantially improved by this PR; CI (no `-shuffle`) is unaffected. A one-line `resetFlags(t)` would close it.

Approving.

PR #67, opened 2026-05-29
Warn on trusted-proxy principal overwrite + bearer+proxy test coverage

cpcloud:
## Summary

Follow-up to #66. Two related changes:

**Production:** when bearer identity (static token, bootstrap, or DB token) and trusted-proxy mode are both configured and a request lands on a trusted listener, the trusted-proxy middleware silently overwrites the bearer-derived principal. The silent overwrite is documented behavior but invisible to an operator who configured both modes on the same listener. `withTrustedProxyActor` now emits a `slog.Warn` per request with structured fields (`listener`, `upstream_kind`, `upstream_actor`, `header`) so the overwrite is observable in logs. Single-mode deployments stay silent.

`ServerConfig` grows a `Logger *slog.Logger` field defaulting to `slog.Default()`; tests inject a per-test capture without touching the global default.

**Test:** the original trusted-proxy e2e tests built a server with **only** the proxy config — no `Auth.Token`, no `RequireTokenIdentity` — so the `requireBearer → withTrustedProxyActor` chain was never exercised together. Adds a sibling helper (`startBearerProxyTestServer`) that wires both layers and captures the slog output, plus four tests for the four interesting interactions:

- **Static bearer + header on trusted listener** — header wins, body actor ignored. Asserts the overwrite warning fires with `upstream_kind=static_token`.
- **Static bearer + missing header on trusted listener** — `400 actor_header_required` (not 401, since bearer auth succeeds).
- **Wrong bearer + header on trusted listener** — `403 auth_invalid` from `requireBearer` before the proxy middleware can run. Locks the middleware order.
- **Identity-mode DB token + conflicting header** — mints a DB token for `bob`, sends a close with `X-Kata-Actor: alice`. The header wins; the DB-token identity for `bob` is silently discarded. Asserts the warning fires with `upstream_kind=db_token`.

Three commits, total ~290 lines: feat (warning), test (bearer+proxy helper + four tests + warning assertions), small comment cleanup.

## Test plan

- [x] `go test ./internal/daemon/` — all pass
- [x] `go test ./internal/daemon/ -race` — race-clean (the slog capture uses a sync-mutex buffer)
- [x] `golangci-lint run ./internal/daemon/...` — clean

### DIFF
--- a/internal/daemon/trusted_actor_e2e_test.go
+++ b/internal/daemon/trusted_actor_e2e_test.go
@@ -1,6 +1,7 @@
 package daemon_test
 
 import (
+	"context"
 	"encoding/json"
 	"net"
 	"net/http"
@@ -13,6 +14,7 @@ import (
 
 	"go.kenn.io/kata/internal/config"
 	"go.kenn.io/kata/internal/daemon"
+	"go.kenn.io/kata/internal/db"
 )
 
 // startTrustedProxyTestServer pre-binds a loopback TCP listener, then builds a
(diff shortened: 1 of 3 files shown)


mariusvniekerk (approved):

internal/daemon/trusted_actor_e2e_test.go
```diff
@@ -301,3 +349,164 @@ func TestTrustedProxyHeader_TUIBypassRequiresCloseValidation(t *testing.T) {
...
+		"/api/v1/projects/"+strconv.FormatInt(projectID, 10)+"/issues/"+issueRef+"/actions/close",
+		body, map[string]string{
+			"Authorization": "Bearer wrong-bearer",
+			"X-Kata-Actor":  "proxy-user",
+		})
+	assertAPIError(t, resp.StatusCode, raw, http.StatusForbidden, "auth_invalid")
+}
+
+// TestTrustedProxyHeader_DBTokenIdentityHeaderOverwrites exercises the case
+// the spec §3.4 precedence table explicitly calls out: identity-mode bearer
```
mariusvniekerk: i'd avoid encoding paragraphs into the doctext here, since this doesn't even call out the doc

mariusvniekerk (changes requested):

internal/daemon/trusted_actor_e2e_test.go
cpcloud: Should be fixed now.

internal/daemon/trusted_actor_e2e_test.go
```diff
@@ -301,3 +349,163 @@ func TestTrustedProxyHeader_TUIBypassRequiresCloseValidation(t *testing.T) {
...
 	assertAPIError(t, resp.StatusCode, raw, http.StatusBadRequest, "validation")
 }
+
+// TestTrustedProxyHeader_StaticBearerHeaderWins exercises the simplest
+// stacked case: a daemon with both bearer auth (static token) and trusted-
+// proxy mode configured. A request that supplies a valid bearer AND the
+// trusted-proxy header on the trusted listener must succeed, with the header
+// value credited as the actor (not the body actor, and not anything derived
+// from the bearer — static tokens have no associated actor).
+func TestTrustedProxyHeader_StaticBearerHeaderWins(t *testing.T) {
```
mariusvniekerk, line 387: this workflow should warn the user about this case, we should be able to detect this via configuring the slog here and observing the logs.

internal/daemon/trusted_actor_e2e_test.go
```diff
@@ -301,3 +349,163 @@ func TestTrustedProxyHeader_TUIBypassRequiresCloseValidation(t *testing.T) {
...
+
+// TestTrustedProxyHeader_DBTokenIdentityHeaderOverwrites exercises the
+// stacked case where bearer identity AND trusted-proxy attribution are both
+// configured. requireIdentityBearer admits the DB token and sets
+// PrincipalDBToken{Actor: "bob"}; the trusted-proxy middleware then
+// overwrites that principal with PrincipalTrustedProxy{Actor: "alice"} on the
+// trusted listener. The resulting event must credit "alice" (header), not
+// "bob" (token), not "client-claim" (body). Pin this so a future change that
+// reverses the precedence (or silently merges the two principals) fails here.
+func TestTrustedProxyHeader_DBTokenIdentityHeaderOverwrites(t *testing.T) {
```
mariusvniekerk, line 498: This should have the same warning when both are configured.

mariusvniekerk (changes requested):

internal/daemon/trusted_actor_e2e_test.go
cpcloud: Added in 3eeb5ee (warning) + a1a281e (test asserts it). `withTrustedProxyActor` now logs a slog `Warn` per request when it overwrites an upstream bearer-derived principal (static_token / bootstrap / db_token), with structured fields (listener, upstream_kind, upstream_actor, header). The test captures the log via a per-test `slog.Logger` and asserts `upstream_kind=static_token` appears.

internal/daemon/trusted_actor_e2e_test.go
cpcloud: Same warning landed in 3eeb5ee; this test now asserts `upstream_kind=db_token` shows up in the captured log alongside the actor-credit assertion.

mariusvniekerk (approved):

PR #69, opened 2026-05-30
Harden JSONL export and import safety

wesm:
## Summary
- remove the JSONL decoder scanner token cap so large issue bodies round-trip
- export from a single read-only snapshot and replace export files only after temp-file sync/close succeeds
- import into a temp DB and install it only after replay succeeds, preserving existing targets on failed forced imports
- audit mutation no-op paths so commit errors are checked before returning result values
- address review #20995 by removing the Windows delete-and-retry export fallback

## Verification
- `go test ./... -count=1`

### DIFF
--- a/cmd/kata/export.go
+++ b/cmd/kata/export.go
@@ -111,7 +111,7 @@ func writeExportOutput(ctx context.Context, d *db.DB, output string, opts jsonl.
 }
 
 func replaceExportOutput(tmpName, output string) error {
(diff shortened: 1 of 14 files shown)


mariusvniekerk:
roborev, you don't know what compiling means.

mariusvniekerk (approved):

wesm:
stale finding. Merging

PR #74, opened 2026-05-31
fix(daemon): detach auto-started daemon from caller console/stdio on Windows

### LINKED ISSUE #73: Windows: daemon-spawned git fails with "exit status 0xc0000142" once the launching shell exits
## Symptom

On Windows, after a kata daemon has been running for a while, every command that reaches the daemon fails:

```
PS C:\src\some-repo> kata list
kata: git remote: exit status 0xc0000142
```

`0xc0000142` is the Windows status code `STATUS_DLL_INIT_FAILED` — `git` is failing to *initialize as a process*, not failing logically.

## Root cause

`git remote` is run **inside the daemon** (`internal/daemon/handlers_projects.go` → `config.ComputeAliasIdentity` → `readGitRemote` → `runGit`), to resolve a workspace's project identity. The `kata:` prefix is the CLI relaying the daemon's error.

The auto-started daemon is spawned by `daemonclient.autoStart`. On Windows `detachChild` is a **no-op**, so the daemon is created **sharing the console of whatever shell first launched it**. When that launching `kata` process exits and its console is torn down, the long-lived daemon is left holding a handle to a **destroyed console**. Git for Windows is built on the msys2 runtime, which requires a usable console to initialize; spawned against a dead console it aborts at startup with `STATUS_DLL_INIT_FAILED` (`0xc0000142`). Every subsequent daemon command that resolves identity then fails.

## Reproduction

1. Run any `kata` command that auto-starts a daemon (e.g. `kata list`).
2. Close / exit the shell that launched it (its console is destroyed).
3. From a new shell, run `kata list` again → `git remote: exit status 0xc0000142`.

Confirmed against a daemon that had been running for three days; `kata daemon stop` followed by a fresh `kata list` (daemon re-attached to a live console) works until that launcher exits again.

## Fix

Detach the auto-started daemon properly so it never depends on the caller's console or stdio:

- Create the daemon with `DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP` (the Windows analogue of the Unix `Setpgid`). With no inherited console, Windows allocates a fresh console for each `git` child and it initializes normally.
- Route the auto-started daemon's stdout/stderr to a `daemon.log` file instead of inheriting the caller's stderr. A now-correctly-surviving daemon that holds the caller's stderr pipe hangs any parent that captures output (command substitution, CI, pipelines); this avoids that and persists daemon logs.

PR to follow.

njt:
Fixes #73.

## Problem

On Windows, once a kata daemon has outlived the shell that launched it, every command that reaches the daemon fails with:

```
kata: git remote: exit status 0xc0000142
```

`git remote` runs **inside the daemon** (`handlers_projects.go` → `ComputeAliasIdentity` → `readGitRemote`). `0xc0000142` is `STATUS_DLL_INIT_FAILED` — git failing to initialize as a process.

The auto-started daemon (`daemonclient.autoStart`) is created **sharing the launching shell's console**, because `detachChild` is a no-op on Windows. When that shell exits and its console is destroyed, the long-lived daemon holds a dead console handle. Git for Windows' msys2 runtime needs a usable console to initialize, so it aborts at startup.

## Fix

1. **`detach_windows.go`** — create the daemon with `DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP` (the Windows analogue of the Unix `Setpgid`). With no inherited console the daemon never depends on the caller's console lifetime, and Windows allocates a fresh console for each `git` child so it initializes normally.

2. **`ensure.go`** — route the auto-started daemon's stdout/stderr to a `daemon.log` file under the data dir instead of inheriting the caller's stderr. Once the daemon correctly *survives* its launcher (point 1), a retained stderr pipe would hang any parent that captures output (command substitution, CI, pipelines); this avoids that and persists the daemon's log. Falls back to the null device if the log can't be opened — it never hands the daemon the caller's stderr.

Unix behavior is unchanged (`detach_windows.go` is Windows-only; on Unix the daemon already detaches via `Setpgid`, and it now logs to `daemon.log` instead of the caller's terminal on auto-start).

## Testing

- `go build ./...`, `go vet ./...`, `go test ./internal/daemonclient/` all pass.
- New Windows-only regression test (`detach_windows_test.go`) pins that `detachChild` sets `DETACHED_PROCESS` and preserves pre-existing creation flags.
- Manual, on Windows 11 with Git for Windows 2.53:
  - Reproduced the original failure against a 3-day-old daemon.
  - With the fix: a fresh `kata list` cold-starts the daemon and returns cleanly (no hang); a `kata list` from a *separate* shell — after the launching console is gone — still resolves `git remote` successfully (no `0xc0000142`); `daemon.log` is created in the data dir.

### DIFF
--- a/internal/daemonclient/detach_windows.go
+++ b/internal/daemonclient/detach_windows.go
@@ -2,9 +2,37 @@
 
 package daemonclient
 
-import "os/exec"
+import (
+	"os/exec"
+	"syscall"
+)
 
-// detachChild is a no-op on Windows; the kata daemon does not officially
-// support Windows in Plan 1 (Unix sockets only), but the build must still
-// compile for go test ./... on Windows CI.
-func detachChild(_ *exec.Cmd) {}
+// DETACHED_PROCESS is not exported by the syscall package; its value is
+// fixed in winbase.h. A process created with this flag has no console.
+const detachedProcess = 0x00000008
+
+// detachChild fully decouples the auto-started daemon from the launching
+// shell's console.
+//
+// On Windows a child created without special flags shares its parent's
+// console. When that console is later torn down — the foreground `kata`
+// process that auto-started the daemon exits — the long-lived daemon is left
+// holding a handle to a destroyed console. Any console subprocess it then
+// spawns fails at process startup: Git for Windows in particular is built on
+// the msys2 runtime, which requires a usable console to initialize, so it
+// aborts with STATUS_DLL_INIT_FAILED (reported by Go as "exit status
+// 0xc0000142"). The daemon resolves project identity by running `git remote`,
+// so every command that reaches the daemon then fails.
+//
+// DETACHED_PROCESS gives the daemon no console at all, so it never depends on
+// the launcher's console lifetime. When the daemon later spawns git, Windows
+// allocates a fresh console for that child and git initializes normally. This
+// mirrors the Unix detachChild, which uses Setpgid to isolate the daemon from
+// the foreground process group. CREATE_NEW_PROCESS_GROUP additionally ensures
+// a Ctrl+C / Ctrl+Break aimed at the caller is never delivered to the daemon.
+func detachChild(cmd *exec.Cmd) {
+	if cmd.SysProcAttr == nil {
+		cmd.SysProcAttr = &syscall.SysProcAttr{}
+	}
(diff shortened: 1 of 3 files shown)


njt:
Superseded by #75, which folds this detach fix into the full Windows daemon port (and lands it on a green Windows test suite). Closing in favor of that PR.

PR #75, opened 2026-05-31
Port the kata daemon to Windows

njt:
Ports the kata daemon to Windows. `go build`, `go vet`, `go test`, and `go mod tidy` are clean on Windows and unchanged on Unix.

Supersedes #73 / #74 (the `0xc0000142` detach fix is folded in as its own commit).

## What's in it

**Daemon port** — platform-specific pieces behind build tags:
- **Signaling** — Windows has no cross-process SIGTERM/SIGHUP, so stop/reload use per-daemon named events (`Local\kata-stop|reload-<dbhash>-<pid>`) set by `daemon stop`/`reload`. Unix keeps SIGTERM/SIGHUP.
- **ProcessAlive** — Windows uses `OpenProcess` + `GetExitCodeProcess(STILL_ACTIVE)`; Unix keeps `kill(0)`.
- **Namespace** — no uid on Windows, so the per-user socket-dir tag derives from a sanitized `USERNAME`.

**Console detach fix** (the original bug) — the daemon runs `git remote` to resolve project identity. On Windows the auto-started daemon shared the caller's console; once that shell exited, git aborted at startup with `STATUS_DLL_INIT_FAILED` (`exit status 0xc0000142`). Now the daemon is created with `DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP` and routes stdout/stderr to `daemon.log` instead of the caller's stderr (which a surviving detached daemon would hold open, hanging output-capturing parents).

**Test portability** — stock `main` had ~25 Windows failures in a Unix-first suite:
- `token_identity` e2e test tagged `!windows` like its siblings.
- hookprobe helper built as `hookprobe.exe`; hook-config tests use bare-name / per-OS absolute commands.
- Unix 0600 file-mode assertions skipped on Windows (ACL-based there).
- `shouldRefuseAutoStartDaemon` normalized via `filepath.ToSlash` (a real bug: it looked for `\go-build` on Windows).
- agent-output tests assert through `agentValue`; beads-from-live-bd tests skip on Windows; stop/reload tests stand up the named events a real daemon would.

**CI** — adds a `windows` job to `.github/workflows/test.yml` running `go build`, `go vet`, and `go test` on `windows-latest`. The existing ubuntu lanes never compile the `*_windows.go` files or `!windows`-tagged tests, so this is the regression gate for the port. No cgo, so no extra toolchain setup.


wesm:
Thanks. If you don't mind I can take it from here and I'll add CI so this doesn't regress?

wesm:
hallucinated

PR #83, opened 2026-06-01
Adopt kit git helpers

mariusvniekerk:
## Summary
- Switch git remote discovery to `go.kenn.io/kit v0.1.0` helpers instead of the local subprocess wrapper.
- Thread `context.Context` through project name and alias derivation so canceled requests stop before git work continues.
- Update CLI, TUI, and daemon call sites to pass context through the resolution flow.

## Testing
- `go test ./...`
- Added coverage for canceled-context handling in project identity derivation.


wesm:
Can any code be deleted as a result of this?

mariusvniekerk:
think the main reduction for kata probably more comes from from of the other parts of kit

PR #86, opened 2026-06-01
feat(init): add --with-agents to write kata guidance into AGENTS.md

mjacobs:
## Summary

`kata init` writes only `.kata.toml` and a `.gitignore` line, so a freshly bound workspace gives coding agents no in-repo hint that kata is the issue ledger — agents have to be told to run `kata quickstart` out of band. Beads, by contrast, writes/updates `AGENTS.md` on `bd init` by default.

This adds an **opt-in** `--with-agents` flag to `kata init` that writes a marker-delimited kata briefing into `AGENTS.md` in the workspace, pointing agents at `kata quickstart` and the close discipline.

Design choices, aimed at fitting kata's "smallest repo footprint" philosophy:

- **Off by default.** A plain `kata init` still writes only `.kata.toml`; nothing changes unless you ask for the block.
- **Marker-delimited + idempotent.** The block lives between `<!-- BEGIN KATA ... -->` / `<!-- END KATA -->`. Re-running refreshes only kata's section, preserves everything else in the file, appends cleanly when `AGENTS.md` already exists, and errors rather than guessing when a begin marker has no matching end.
- **Non-destructive beads migration.** If `AGENTS.md` (or a real, non-symlinked `CLAUDE.md`) still carries a Beads integration block, kata refuses to edit it in place. It leaves the original byte-for-byte untouched and writes a `<file>.kata-proposed` sidecar with the Beads block removed and kata's block added, then prints how to adopt (`mv`) or discard it. A symlinked `CLAUDE.md` is skipped so kata never rewrites a file that just points at `AGENTS.md`.
- **Short by design.** The injected block mirrors the quickstart contract in brief and points back at `kata quickstart` rather than duplicating it.
- **Works in both init flows** (path-free client-side write and path-based fallback), writing alongside where `.gitignore` is managed.

kata's own repo treats `AGENTS.md` as the source of truth with `CLAUDE.md` as a symlink (per #57), so `CLAUDE.md` is only ever touched in the beads-conflict case above, and never when it is a symlink.

## Changes

- `cmd/kata/init.go`: `--with-agents` flag; `applyAgentGuidance` orchestrator; `ensureAgentsBlock`/`upsertKataBlock` managed-block handling; `writeMigrationProposal` + beads detection (`hasBeadsBlock`/`stripBeadsBlock`/`regularFileWithBeads`) for the sidecar path; `warnBeadsConflict` user guidance. Wired into `runNameInit` and `runStartPathInit`.
- `cmd/kata/init_test.go`: writes-when-absent, no-op-without-flag, idempotent, appends-to-existing, refreshes-stale-block, plus beads-block-in-AGENTS sidecar, beads-block-in-CLAUDE sidecar, and symlinked-CLAUDE-skipped.
- `cmd/kata/init_with_agents_e2e_test.go`: end-to-end flows via the real CLI — new empty repo, existing CLAUDE.md + AGENTS.md preserved, and a beads import migration.
- `docs/get-started/quickstart.md`, `docs/workflows/agents.md`: document the flag and the sidecar behavior.

Local checks: `go test ./cmd/kata/`, `go vet`, and `golangci-lint` all pass. (`nilaway` not installed in my environment; change is plain string handling with no new nil paths.) Also verified by hand against a real 59-issue beads database: AGENTS.md preserved, CLAUDE.md untouched, sidecar produced as expected.

Opening as **draft** to get maintainer read on whether an opt-in AGENTS.md writer fits kata's direction before polishing.

### DIFF
--- a/cmd/kata/init.go
+++ b/cmd/kata/init.go
@@ -376,20 +376,47 @@ func warnAgentsUpdate(err error) {
 	fmt.Fprintf(os.Stderr, "kata: warning: could not update AGENTS.md: %v\n", err)
 }
 
-// warnBeadsConflict tells the user kata declined to edit file in place because
-// it still carries a beads integration block, and points at the sidecar
+// warnBeadsConflict tells the user kata declined to edit a file in place
+// because it still carries a beads integration block, and points at the sidecar
 // proposal kata wrote instead. Suppressed in machine-output modes, matching
 // warnGitignoreUpdate.
-func warnBeadsConflict(file, sidecar string) {
+func warnBeadsConflict(originalPath, sidecarPath string) {
 	if currentOutputMode() != outputHuman {
 		return
 	}
-	base := filepath.Base(sidecar)
-	fmt.Fprintf(os.Stderr,
+	cwd, err := os.Getwd()
+	if err != nil {
+		cwd = ""
+	}
+	fmt.Fprint(os.Stderr, beadsConflictMessage(cwd, originalPath, sidecarPath))
+}
+
+// beadsConflictMessage builds the human hint for a declined in-place edit.
+// Paths are rendered relative to cwd when they sit under it, else absolute, so
+// the `mv` command is copy-pasteable no matter which directory init was invoked
+// from — the write destination is the workspace root, which need not be cwd.
+func beadsConflictMessage(cwd, originalPath, sidecarPath string) string {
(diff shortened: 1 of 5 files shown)


wesm:
Thanks Matt! I'll get this PR in in the next day or so

wesm:
false positive

PR #87, opened 2026-06-01
feat(db): add storage abstraction and Postgres schema shell

cpcloud:
## Summary

- Adds a backend-neutral `db.Storage` contract, moves the SQLite implementation to `internal/db/sqlitestore`, and routes production SQLite opens through `internal/db/storeopen`. Daemon, federation, and command paths now hold `db.Storage`, with the daemon raw-SQL escape hatches lifted into store methods.
- Adds `config.KataDSN(ctx)` and `[storage].dsn` config resolution with precedence `KATA_DSN > KATA_DB > [storage].dsn > default`, plus credential-safe DSN identity/redaction for runtime files and error paths.
- Adds `internal/db/pgstore` as a Postgres schema/open/bootstrap shell using pgx. Postgres is intentionally incomplete in this PR: pgstore can open/bootstrap a fresh schema directly and satisfies `db.Storage` through generated stubs, but normal `storeopen` daemon/CLI selection rejects `postgres://` and `postgresql://` DSNs until core domain methods land.
- Keeps schema evolution migration-free: fresh SQLite databases bootstrap from canonical `schema.sql`, pre-current SQLite databases upgrade through JSONL cutover before open, and existing version-0 SQLite databases are treated as legacy/cutover candidates rather than fresh files. Direct pgstore schema-version handling remains schema-shell only.
- Moves current-schema JSONL import/export onto the storage boundary: current exports read through `Storage` export iterators with snapshot support when the backend provides it, read-only pre-current SQLite exports route through the legacy version-aware exporter, and `jsonl.Import` decodes/fills records before delegating atomic replay to `Storage.ImportReplay`.
- Preserves federation/claim interop by moving claim and federation methods into `Storage` and porting the SQLite implementations, including `AdoptProjectIntoFederation` helpers and local-first claim semantics from main (#71).
- Cleans up review churn: pgstore stubs are generated by `internal/db/pgstore/stubgen` and drift-tested, sqlitestore export iterators share `streamRows[T]`, JSONL timestamp normalization is consolidated, and SQLite query code remains split by query area.


wesm:
hallucination

Issue #97, opened 2026-06-04
Publish a versioned OpenAPI schema snapshot with documented compatibility


mjacobs:
# Publish a versioned OpenAPI schema snapshot with documented compatibility

I'm building an external read-only viewer for Kata, and wanted to generate a
typed client against the daemon's HTTP API. I found there's no published schema
to generate from — the only option is to hand-copy wire structs out of
`internal/`. This isn't viewer-specific: any out-of-process client that talks to
the daemon over HTTP hits the same wall, which is why I'm raising it as a
general API-contract gap rather than a one-off request.

_(References below are against `kenn-io/kata@3ec9d25`, 2026-06-03; file paths
may drift, but the symbols/behavior are the stable part of the ask.)_

## Problem

Kata uses Huma internally, but the generated OpenAPI document is currently
disabled — `internal/daemon/server.go` sets `humaConfig.OpenAPIPath = ""` ("Plan
1: no /openapi.json"). External clients that want to generate typed clients
therefore have to hand-copy wire structs from `internal/` packages (which are
not importable) and have no way to detect or fail clearly on an incompatible
daemon version.

## Proposal

Publish the daemon's OpenAPI schema as a versioned snapshot with documented
compatibility expectations. This is intentionally framed as a versioned
_snapshot per release_ with stated compatibility guarantees — not a promise of a
forever-stable API.

A versioned schema snapshot becomes the public contract for HTTP DTOs, so
external clients no longer depend on Kata's internal Go structs. This benefits
any generated-client workflow, not a single tool.

- [ ] Expose or export the OpenAPI document (re-enable `OpenAPIPath`, or provide
      a `kata` command / committed artifact that emits it).
- [ ] Commit or publish the schema artifact per release.
- [ ] Add docs or a command for regenerating it.
- [ ] Include daemon/API schema version in health/version output.
- [ ] Document compatibility expectations between client and daemon versions
      (what may change, and how breaking changes are signalled).

mjacobs:
Opened #99 as a first implementation slice for this.

One note on framing, since the issue wording leans on "public contract": I'm treating that PR as a *generated snapshot of the current daemon API* — useful for client generation and review — rather than a compatibility guarantee or a commitment to support external daemon clients. The "documented compatibility" piece here, and surfacing the schema version in `health`, are intentionally left as follow-ups pending your preference on how far you'd want to take this.

[...]

PR #99, opened 2026-06-05
feat(daemon): add kata openapi command + committed OpenAPI schema

mjacobs:
## What

Adds a `kata openapi` command that prints the daemon's OpenAPI 3.1 schema, and
commits the generated artifact at `api/openapi.yaml` with a golden test that
keeps it from drifting.

This is a **reviewable first slice of #97** — a generated *snapshot of the
current daemon API*, useful for client generation and review. It deliberately
does **not** make stronger compatibility promises, and it's meant as an
implementation path for your feedback rather than an assumption about policy.
The "API schema version in `health` output" and "compatibility-expectations doc"
items from #97 are left as follow-ups, pending your preference on how far you
want to take this.

## Approach

- **Runtime `/openapi.json` stays disabled.** The existing decision in
  `internal/daemon/server.go` is untouched; the schema is generated
  **in-process** from the route definitions (no running daemon or DB required),
  so this doesn't reverse it or add a network surface.
- **Stdout-only command.** `kata openapi` writes to stdout with no `-o`/file
  flag, so the shipped binary introduces no path/symlink write surface. The
  committed artifact is produced by `make openapi` (a stdout redirect), not by
  the binary writing files itself.
- **Drift guard.** `make openapi` regenerates `api/openapi.yaml`;
  `TestOpenAPIArtifactUpToDate` fails if the committed file no longer matches the
  routes. Output is deterministic (verified across repeated generation), and
  regenerating after a route/schema change is a single `make openapi`.

## Error-schema fidelity (worth a look)

`APIError` serializes through a custom `MarshalJSON` into the `ErrorEnvelope`
wire shape — `{status, error:{code,message,hint?,data?}}` — which Huma's
reflection can't see. Left alone, the generated schema would describe every
non-2xx response using `APIError`'s raw Go fields (flat, exported names): a
schema that misrepresents all error responses. This PR has `APIError` implement
`huma.SchemaProvider` so the document references the true `ErrorEnvelope` shape
instead.

## Open to your preference

I defaulted to the most concrete thing to review, but I'm happy to adjust any of
these rather than presume the call is made:

- **Committing the artifact vs not** — if a ~5.2k-line generated file is review
  noise, this can be command-only (no `api/openapi.yaml` in-tree), or the
  artifact can be published only at release time instead of committed.
- **Which routes appear** — the snapshot currently includes every registered
  route. If some endpoints shouldn't be presented as part of a published
  surface, I can filter them.
- **The golden test** — it's there to stop silent drift, but easy to relax or
  drop if you'd rather not couple route changes to artifact regeneration.

## Notes

- Verified locally: `go test` (incl. `-short` across `internal/...`/`cmd/...`),
  `go vet`, `golangci-lint` (0 issues), `nilaway` (clean), `gofmt`.

Part of #97.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: build install test test-short test-stress test-federation-docker lint vet clean fmt nilaway tui tui-demo docs-install docs-build docs-serve docs-check docs-deploy docs-screenshots docs-assets-branch
+.PHONY: build install test test-short test-stress test-federation-docker lint vet clean fmt nilaway openapi tui tui-demo docs-install docs-build docs-serve docs-check docs-deploy docs-screenshots docs-assets-branch
 
 GOFLAGS_TEST := -shuffle=on
 GOBIN ?= $(HOME)/.local/bin
@@ -15,6 +15,11 @@ install:
 
 test:
 	go test $(GOFLAGS_TEST) ./...
+
+# Regenerate the committed OpenAPI schema artifact from the daemon's routes.
+# The TestOpenAPIArtifactUpToDate test fails if api/openapi.yaml drifts from this.
+openapi:
+	go run ./cmd/kata openapi > api/openapi.yaml
 
 test-short:
 	go test -short $(GOFLAGS_TEST) ./...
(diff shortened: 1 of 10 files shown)


wesm:
I can address this roborev finding

PR #101, opened 2026-06-05
fix(daemon): publish unix runtime addresses as URLs

cpcloud:
Fix local daemon runtime discovery interop by publishing URL-shaped Unix socket addresses in daemon runtime records.

Kata already uses `go.kenn.io/kit/daemon.Endpoint.ConfigAddress()` as the stable config/runtime address form. The startup path passed the bound endpoint into `NewRuntimeRecord`, which stores `Endpoint.Address` directly. For Unix endpoints that produced a bare absolute socket path in `daemon.<pid>.json`; TCP endpoints remained usable because their dial address and config address are both `host:port`.

This matters for Kataflow and similar runtime-record consumers that accept `unix://...`, `http://...`, `https://...`, or bare `host:port` endpoint strings, but intentionally do not treat arbitrary absolute filesystem paths as daemon addresses. With this change, Unix records publish `unix:///path/to/daemon.sock` while TCP records continue to publish `127.0.0.1:<port>`.

The new daemon-start coverage reads the raw runtime JSON before shutdown and asserts both the Unix URL form and the existing TCP host:port contract.

### DIFF
--- a/cmd/kata/daemon_cmd.go
+++ b/cmd/kata/daemon_cmd.go
@@ -380,7 +380,9 @@ func runDaemonWithListen(ctx context.Context, listen string, insecureReadonly bo
 	}
 	defer func() { _ = listener.Close() }()
 
-	rec := kitdaemon.NewRuntimeRecord("kata", version.Version, runtimeEndpointForListener(endpoint, listener))
+	runtimeEndpoint := runtimeEndpointForListener(endpoint, listener)
+	rec := kitdaemon.NewRuntimeRecord("kata", version.Version, runtimeEndpoint)
+	rec.Address = runtimeEndpoint.ConfigAddress()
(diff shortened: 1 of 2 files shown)


cpcloud:
Roborev job 5708 was a false positive. The review assumed the new runtime JSON address would flow through consumers as `unix://unix:///path` or get passed to `UnixTransport` with the scheme still attached.

That is not how the current kit API behaves: `kitdaemon.RuntimeRecord.Endpoint()` strips a leading `unix://` from `RuntimeRecord.Address` before returning `Endpoint.Address`. As a result, existing consumers that call `r.Endpoint().ConfigAddress()` still get exactly one `unix://` prefix, and the code path that passes `ep.Address` to `UnixTransport` still receives the bare socket path.

I verified the focused daemon/client/runtime checks after closing the roborev finding.

PR #107, opened 2026-06-09
Support named daemon catalog

cpcloud:
## What changed

This adds a named daemon catalog flow for repo-specific daemon selection without putting secrets in repo-local config. `.kata.toml` and `.kata.local.toml` can select `[server].daemon = "work"`, and each user's `KATA_HOME/config.toml` resolves that name through a `[[daemon]]` catalog entry.

Client auth now treats the selected daemon's catalog entry as the auth boundary: `token_env` is the preferred secret path, explicit `KATA_AUTH_TOKEN` and explicit bearer clients still override token lookup, and unrelated global auth is not consumed for a selected named daemon.

Named local daemons get separate runtime/storage identity via `kata.<name>.db`, and `kata daemon start --name` starts only local catalog entries. Named starts do not inherit the default daemon listener or hosted `PORT` fallback unless the operator supplies `--listen`.

## Why

The repo-local files should be selectors, not secret stores. This lets a repository say "use the work daemon" while keeping each developer's URL/token/env details in their own global config and shell environment.

## Tradeoffs

This intentionally uses a single global daemon catalog so daemon selection, auth env var names, and local/remote targets have one owner. Inline catalog `token` remains supported, but the docs steer users to `token_env`.

## Where to look

- `internal/client/remote.go` and `internal/client/named_daemon.go` for workspace daemon selection.
- `internal/client/auth.go` for selected-daemon auth isolation and explicit-token precedence.
- `cmd/kata/daemon_cmd.go` and `internal/daemon/namespace.go` for named local daemon startup/runtime behavior.
- `docs/reference/configuration.md` and `docs/operations/remote-daemon.md` for the documented setup.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -89,9 +89,12 @@ A daemon can serve clients on other hosts over a private network:
   `kata daemon start --listen 100.64.0.5:7777`. Set
   `listen = "100.64.0.5:7777"` in `<KATA_HOME>/config.toml` so every
   daemon (including the auto-started one) binds TCP.
-- Client: `export KATA_SERVER=http://100.64.0.5:7777` or commit a
-  gitignored `.kata.local.toml` with `[server] url = "..."` next to
-  `.kata.toml`. `KATA_SERVER` env wins.
+- Client: `export KATA_SERVER=http://100.64.0.5:7777` for one-off routing, or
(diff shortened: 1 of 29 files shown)


cpcloud:
This is not what I originally had in mind after poking around the details, closing.

Issue #104, opened 2026-06-07
Feature: move/reassign an issue to a different project (kata move / kata edit --project)


Joi:
## Problem

`kata edit` can change an issue's title/body/owner/priority/parent/labels/relationships, but **not its project**. `projects merge` only operates on whole projects. So a *single* mis-filed issue can't be relocated without close-and-recreate — which loses history, comments, short_id, and relationships.

## Where it bites (real usage)

Running one daemon with ~16 projects (repo-per-project). Two recurring needs with no clean path:

1. Issues filed into a pilot-era catch-all project before per-repo projects existed.
2. Rationalizing an overloaded project (107 open issues) by moving subsystem issues into dedicated / inbox projects.

We fell back to `area:*` labels for navigation — helpful, but labels don't fix *mis-filed* issues.

## Proposal

`kata move <ref> <project>` (or `kata edit <ref> --project <name>`), with `--dry-run`.

## Design notes

- `short_id` is per-project; a move must re-key `qualified_id` while keeping the ULID stable — decide preserve-vs-reissue `short_id` and collision handling in the target project.
- Preserve comments / events / labels / owner / priority / timestamps; emit a `moved` event (old → new project) for audit.
- Cross-project relationships: the CLI already accepts `project#short_id` refs, so `blocked_by` / `blocks` / `parent` / `related` should survive a move — worth confirming.
- Nice-to-have: bulk move by label/filter (the "split an overloaded project" case directly).

Version: `33cfc4b`

mariusvniekerk:
Good feature. Clanking.

wesm:
This is done now

PR #111, opened 2026-06-10
fix(links): make cross-project link errors honest and document the limitation

Joi:
## What this fixes

Found while using `kata move` (#106 / #104) to reorganize issues between projects: every error around the cross-project link constraint points the user at a workaround that cannot work, ending in a dead end with severed links.

Cross-project links are intentionally unsupported — the schema's `trg_links_same_project_insert`/`_update` triggers abort any link row whose endpoints live in different projects. The messaging just never says so:

### Repro (two projects `alpha`, `beta`; `alpha#z0bv --related alpha#0qxz`)

```
$ kata move alpha#z0bv beta
ERR move conflict: cannot move: 1 cross-project link(s) anchored in source project
```
The links aren't cross-project (they can't be); they *would become* cross-project. The wording suggests detach → move → re-attach works. So:

```
$ kata edit alpha#z0bv --remove-related 0qxz   # detach
$ kata move alpha#z0bv beta                    # moves fine
$ kata edit beta#z0bv --related "alpha#0qxz"   # re-attach?
ERR edit validation: --related: cross-project refs not supported here; pass the issue's ULID instead of "alpha#0qxz"
```
The hint says a ULID should work. It can't — `resolveIssueRef` (deliberately, to prevent cross-project fishing via the URL path) rejects ULIDs outside the URL issue's project, and the schema trigger would refuse the row anyway:

```
$ kata edit beta#z0bv --related 01KTSW6M314NZSVH5MTDEY0QXZ
ERR edit not_found: issue not found
```

So the user is told to do something impossible, and the failure is indistinguishable from a typo'd ref.

## Changes

Given the schema triggers, I treated "no cross-project links" as the intended design and made the messaging/docs honest rather than changing link semantics:

- **CLI** (`cmd/kata/issue_ref_cli.go`): qualified-ref rejection now states the real constraint — `--related: cross-project links are not supported; "alpha#0qxz" must name an issue in project "beta"` — instead of recommending the ULID dead end.
- **daemon** (`internal/daemon/resolver.go`): link-target resolution misses (links_delta + create-time links) now read `link target not found in this project (links can only join issues in the same project)`. The message is identical for foreign and nonexistent targets, so it leaks nothing about foreign issues (the anti-fishing 404 behavior is unchanged — only the wording).
- **db** (`internal/db/errors.go`): `CrossProjectLinksError` → `cannot move: N link(s) would become cross-project; unlink before moving (links can only join issues in the same project)`.
- **docs + `kata move --help`**: document that `move` refuses while links exist and describe the only sequence that works: remove links → move the issue (and any peers you want to stay connected) → re-link in the target project.

Tests: updated `TestEdit_CrossProjectLinkRefRejected` (now also pins that the error no longer suggests ULIDs), added `TestEditIssue_LinkTargetULIDInOtherProject_404` (foreign-ULID link target 404s with the same-project wording, and the message is byte-identical for a nonexistent ULID), and pinned the reworded 409 in `TestMoveIssue_CrossProjectLinks_409_WithBlockers`. `go test ./...` passes.

## Design question (out of scope here, but worth a maintainer call)

Is "links never span projects" permanent design, or should `move` eventually preserve relationships? Two possible follow-ups if there's appetite:

1. `kata move --with-links` that refuses unless all peers are moved in the same operation (move the closed set), or
2. actual cross-project link support (drop the triggers, decide how `links.project_id` should be scoped, what events/export/federation do with links whose endpoints live in two projects).

Either is a schema/design decision, so this PR only stops the tooling from pointing at a workaround that doesn't exist.

### DIFF
--- a/cmd/kata/edit_test.go
+++ b/cmd/kata/edit_test.go
@@ -253,9 +253,10 @@ func TestEdit_EquivalentParentRefsAccepted(t *testing.T) {
 
 // TestEdit_CrossProjectLinkRefRejected pins that link flags refuse a
 // qualified ref naming a project other than the URL issue's project.
-// The daemon's wire shape (RefForAPI is just the short_id resolved
-// against the current project) would silently target the wrong issue;
-// the user must pass the peer's ULID for a cross-project link.
+// Cross-project links are not supported (the schema forbids them), so
+// the error must state that constraint — and must NOT steer the user
+// toward passing a ULID, which the daemon also resolves only inside the
+// URL issue's project.
 func TestEdit_CrossProjectLinkRefRejected(t *testing.T) {
 	env, dir := setupCLIEnv(t)
 	pid := resolvePIDViaHTTP(t, env.URL, dir)
@@ -264,8 +265,10 @@ func TestEdit_CrossProjectLinkRefRejected(t *testing.T) {
 
 	_, err := runCLICapture(t, env, dir, "edit", subject, "--blocks", "other#"+peer)
 	require.Error(t, err, "cross-project qualified ref must be rejected")
-	assert.Contains(t, err.Error(), "cross-project")
+	assert.Contains(t, err.Error(), "cross-project links are not supported")
 	assert.Contains(t, err.Error(), "--blocks")
+	assert.NotContains(t, err.Error(), "ULID",
+		"error must not suggest ULIDs: the daemon scopes ULID link refs to the issue's project too")
 }
 
 // TestEdit_ConflictDetectedAcrossRefForms pins that the add/remove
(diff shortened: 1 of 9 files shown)


wesm:
Thank you, will take a closer look at making this more ergonomic!

wesm:
I'm working on a complete solution for this, will push changes to this PR and update the description with the details

wesm:
Superseded by #112, which builds on this branch (your commit ships in its history unsquashed) and removes the same-project limitation instead of documenting it: links are now project-independent edges, kata move preserves them, and link-target refs accept bare, qualified, and ULID forms. The design question this PR raised is answered in docs/superpowers/specs/2026-06-10-cross-project-links-design.md on that branch.

PR #112, opened 2026-06-11
feat: links may span projects and survive kata move

wesm:
Supersedes #111. That PR made the same-project limitation honest; this one removes the limitation. It builds on #111's branch — joi's commit ships in this history unsquashed — so the error-message work lands here too, rewritten for the new behavior.

## What changes

A link (`parent`, `blocks`, `related`) is now a project-independent edge between two issues in the same daemon database. Project context is derived from the endpoints, never stored on the edge.

**Storage (v15).** The same-project triggers and `links.project_id` are gone; UID-consistency triggers, uniqueness, one-parent-per-child, and `related`'s row-id canonicalization remain. v14 databases upgrade through the existing JSONL cutover (no data transform — v14 cannot contain cross-project rows). `kata move` no longer gates on links; edges survive untouched because endpoints are row ids and UIDs, both move-stable.

**Daemon.** Link-target refs resolve in three forms: bare `short_id` (subject issue's project), `project#short_id` (named project), 26-char ULID (global). Misses are a plain `issue_not_found` — the old "links can only join issues in the same project" wording is gone. The URL-subject resolver keeps its anti-fishing scoping unchanged. Adds targeting an archived project get `409 link_target_archived` with an unarchive hint; removes tolerate archived projects and soft-deleted peers so existing links can always be cleaned up. Federated claim gates evaluate each endpoint against its own project on every link path, including create-time initial links. POST `/links` parent creation runs the same in-tx cycle guard as the edit path, plus a pre-flight check on `replace=true` so a rejected replace no longer unlinks the old parent first; both ancestor walks share one depth cap (`db.MaxParentDepth`).

**Wire (API 0.2.0).** `LinkPeer` carries always-populated `project` and `qualified_id` (`short_id` stays bare). `IssueOut.parent_short_id` is replaced by a full `parent` peer; `LinkOut.project_id` is removed. Edit responses' `changes` peers carry the same fields, captured inside the mutating transaction so they cannot race a concurrent move. OpenAPI and `pkg/client` are regenerated; the generated client models optional object-valued response fields (`parent`, `changes`, `child_counts`) as pointers, with the 3.0 generator flavor of the schema leaving response `additionalProperties` unset to make that possible. `/health`'s `api_schema_version` is the compatibility signal.

**CLI.** `kata create`/`kata edit` link flags accept all three ref forms; foreign qualified refs are forwarded instead of pre-rejected. `kata show` and `kata edit`'s change summary render foreign peers as `project#short_id` and same-project peers bare.

**TUI.** Nested views key parent grouping by issue UID instead of `{project, short_id}`, so a cross-project parent cannot capture an unrelated same-short-id issue; a parent outside the fetched list renders its children as top-level rows.

**Export/import.** A project-scoped envelope can carry a link edge whose peer issue is absent: import skips those links and any import-mapping records referencing them (aggregate stderr notes), and a later fuller envelope self-heals the edges via natural-key dedupe (`related` by unordered UID pair). Project-filtered event exports scrub omitted-peer refs.

**Federation.** Cross-project links do not propagate through federation: link replication is scoped to a project binding, so an edge materializes remotely only when both endpoints are issues of the same federated project, and the hub rejects push batches whose link payloads reference issues outside the target project. A cross-project link mutation on a push-enabled spoke, or an adoption snapshot embedding one, can land that push batch in quarantine. The federation operations doc states this; lifting the limitation is part of the tracked next stage (kata `nkvn`).

**Events are unchanged in this PR.** A cross-project link mutation emits its event in the subject issue's project only; `issue.links_changed` payload bytes are golden-pinned. Per-edge mirroring into the peer project, edge-keyed sibling-close guards with project-qualified payloads, and federation withholding tests are the tracked next stage (kata `nkvn`); the docs state today's behavior plainly.

## Docs

`docs/reference/cli.md`, `docs/guide/concepts.md`, `docs/reference/http-api.md` (0.2.0 changelog), `docs/operations/federation.md`, and `docs/operations/backup-restore.md`.


wesm:
Both Medium findings from the combined review are fixed:

- **internal/tui/queue_rows.go** — queue hierarchy now resolves parent peers by UID through an index of the loaded issues (0047ddc). A cross-project parent nests its child correctly; a same-short-id issue in the child's project can no longer capture it; an unfetched parent leaves the child top-level. Pinned by `TestBuildQueueRows_CrossProjectParentNestsByUID`.
- **pkg/client/generated/types.go** — optional object-valued response fields (`parent`, `changes`, `child_counts`, `parent_set`, `parent_removed`) are generated as pointers with nil-guarded validation (25398d3). The root cause was the response `additionalProperties: true` relaxation: oapi-codegen-dd emits an optional $ref property as a value type whenever the target schema carries an explicit additionalProperties constraint. The 3.0 generator-input flavor now leaves it unset (same permissive semantics); the published 3.1 schema is byte-identical. Pinned by `pkg/client/optional_object_fields_test.go`.

Also in this push: the parent-replace pre-flight and the storage cycle guard share one depth cap (bedcfe2), and the federation docs were narrowed to the verified cross-project link behavior (3ec2c89).

wesm:
Both Medium findings from the combined review (`1db5ce1`) are fixed in 80f506a:

- **Event hydration** — the subject and related-issue joins in `EventsAfter`, `EventsInWindow`, `eventSelectByID`, and the federation push query now match by row id / UID alone (both globally unique). Cross-project link events hydrate `related_issue_short_id` again, and as the same fix, a moved issue's historical events keep hydrating `issue_short_id` instead of going blank. Pinned by `TestPollEvents_RelatedShortIDForCrossProjectLink` and `TestPollEvents_IssueShortIDSurvivesMove`.
- **Sibling-close guards** — `RecentSiblingCloses` keys the cohort on the parent edge alone; the project filter (and the `projectID` parameter through the storage interface and both guards) is gone, so distributing sibling closes across projects no longer bypasses the close-rate throttle or the repeated-message guard. Pinned by `TestSiblingThrottle_CountsCrossProjectSiblings` (red before: the fourth close returned 200) and `TestRepeatedMessageGuard_CountsCrossProjectSiblings`.

PR #117, opened 2026-06-18
Add global daemon selection flag

wesm:
## Summary
- Add global `--daemon <name>` support for selecting a named daemon catalog entry per invocation.
- Resolve named local and remote targets ahead of ambient remote configuration, including catalog token/token_env auth.
- Update CLI, quickstart, and remote-daemon docs for the new resolution order.


wesm:
looking

PR #120, opened 2026-06-19
Add versioned release workflow

wesm:
kata now has enough release identity plumbing that leaving distribution unversioned makes installs, updates, and changelog history harder to reason about. This PR introduces a repeatable v0.5.0 release path: local release scripts, a documented changelog flow, GitHub release assets, a hosted `https://katatracker.com/install.sh` binary installer, and a kit-backed `kata update` command that uses the same asset-name contract the release publishes.

The release pipeline now follows the sibling CLI pattern with GoReleaser. A semver tag creates the GitHub release from the annotated tag notes, then GoReleaser publishes `amd64` and `arm64` archives for Linux, macOS, and Windows, plus Linux `.deb` and `.rpm` packages and a single `SHA256SUMS` file. Signing is intentionally not part of this first kata release path; update signing can be added later as a separate hardening project.

The installer downloads the latest platform archive from GitHub releases, verifies it against `SHA256SUMS`, and installs `kata` into `/usr/local/bin` or `~/.local/bin`. The docs and README now lead with `curl -fsSL https://katatracker.com/install.sh | bash`, mention the Linux packages, and keep `go install` and manual release downloads documented for source/development installs.

The changelog gives the first release a public documentation home and turns the pre-release commit history into a week-by-week project history with readable calendar ranges. Changelog generation now defaults to deterministic git-log output; command-capable AI generation requires explicit `CHANGELOG_AGENT` opt-in. The Superpowers design and execution artifacts were removed from the branch so repository docs describe shipped behavior rather than agent-local planning notes.


wesm:
Addressed the review and linked CI failure in `4fa5448`:

- `kata update` now requires signed checksum metadata with an embedded Ed25519 public key; `AllowUnsignedChecksums` is gone.
- The release workflow signs each archive checksum and uploads `*.sha256.sig` assets alongside archives and `SHA256SUMS`.
- `kata update` prompts on stderr unless `--yes` is supplied, including JSON/agent modes, and emits JSON/agent install success output.
- The release archive-name test invokes the shell script through `bash`, which fixes the Windows CI failure where `.sh` was executed as a Win32 binary.

The matching private signing material is local at `/tmp/kata-update-signing-key.txt`; set `KATA_UPDATE_SIGNING_PRIVATE_KEY_HEX` from that before cutting `v0.5.0`.

wesm:
Addressed the roborev release-workflow findings in 982292a.

- Tag-triggered `Release` workflow now has read-only contents permission and only builds/uploads artifacts; it no longer reads `KATA_UPDATE_SIGNING_PRIVATE_KEY_HEX`, signs assets, or creates releases.
- New `Publish Release` workflow runs from `workflow_run`, verifies the built tag resolves to the completed run SHA and is contained in `origin/main`, then signs and publishes from that protected workflow path.
- Release creation now uses archive and signature arrays instead of `./kata_*`, so generated `*.sha256.sig` files are uploaded once.

The publisher job uses the `release-signing` environment; configure that environment with required reviewers before cutting the first release.

PR #123, opened 2026-06-23
Use codex for release changelogs

wesm:
The release script should produce the same annotated-tag notes that the rest of the release pipeline publishes. The previous default silently used deterministic git-log output unless an operator explicitly opted into an agent, which made the normal release path easy to run without the expected generated notes.

This makes codex the default release-note generator while preserving an explicit deterministic fallback for tests and manual recovery. Missing or unknown agent choices now fail visibly so a release operator does not accidentally publish lower-quality notes.

The shell test cleanup removes content-grep assertions from the docs checker while keeping build and hydration behavior checks, aligning the suite with the project rule that shell tests should exercise observable behavior rather than implementation strings.


wesm:
not consistent with threat model

PR #134, opened 2026-06-24
Publish release artifacts from tag pushes

wesm:
Release operators expect the local release script to hand off to CI once it pushes an annotated semver tag. The existing tag-triggered workflow only validated that tag, while the actual GitHub release and GoReleaser upload lived behind a separate manual dispatch workflow, making artifact publication easy to miss after a normal release cut.

This moves the publish path into the tag-triggered Release workflow so future releases follow the same tag-push contract as sibling tools. The manual publish workflow remains available for backfills and recovery, but the normal path now creates the GitHub release, re-verifies the immutable tag target, and runs GoReleaser automatically.

A parser-backed release workflow regression test now locks in that the tag-triggered workflow has write permissions, creates the release, and invokes GoReleaser.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,7 +6,7 @@ on:
       - 'v*'
 
 permissions:
-  contents: read
+  contents: write
 
 jobs:
   validate-tag:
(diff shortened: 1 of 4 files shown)


wesm:
accepted tradeoff for now

PR #139, opened 2026-06-27
feat(githubsync): add scoped service credentials and parent-link sync

andy-vdg:
## Summary

- Adds daemon-side GitHub sync credentials for shared deployments: exact `(host, owner)` GitHub App entries, a host-scoped explicit token environment variable, then `gh auth token --hostname <host>` as the local fallback.
- Replaces command-backed repository reads with a scoped HTTP client for REST issue/comment fetches and GraphQL parent discovery; `gh` is now used only as an optional credential source.
- Hardens token egress, Enterprise URL derivation, retry behavior, redaction, and no-credentials error messages so credentials stay bound to the configured GitHub origin, owner, and repository.
- Imports GitHub sub-issue parent links through a paginated GraphQL scan, preserves existing source-managed links when parent data is unsupported or non-authoritative, and keeps locally set parent links from being overwritten.
- Reconciles authoritative parent-scan data for already imported children even when GitHub does not return those children from the incremental issue fetch.
- Adds one-time parent-link backfill state for existing bindings, daemon/API wiring for the new credential config, and regression coverage for backfill state, unsupported GraphQL schema errors, credential routing, parent-only changes, and live parent validation knobs.
- Updates the Zensical docs: operator guidance, disposable validation workflow, design notes, CLI/reference configuration, and HTTP API behavior now describe the service credential path and parent-link sync behavior.

## Why

This PR started as GitHub sub-issue parent import, but the implementation exposed a larger deployment gap: shared or hosted daemons could not rely on ambient `gh` credentials, and parent reconciliation needed explicit authority semantics to avoid deleting hierarchy during transient or unsupported parent-data fetches.

The current branch closes that gap by making credential resolution daemon-owned and origin-scoped, while keeping the transport small and explicit instead of introducing a second GitHub type system. GitHub App credentials are the intended path for multi-owner/shared deployments; host-scoped env tokens and `gh auth token` remain fallback paths for simpler local setups.

## Behavior notes

GitHub-sourced issue fields, comments, and supported parent links remain source-managed. Parent data is authoritative only when the GraphQL scan succeeds for a child; unsupported GitHub Enterprise schemas degrade non-fatally and preserve existing GitHub-sourced parent mappings. Missing parent mappings caused by snapshot skew are skipped and retried on the next sync rather than advancing into destructive reconciliation.

Live validation against a disposable private repository showed that GitHub reparent and unparent operations can leave the child issue `updated_at` unchanged. The runner therefore applies authoritative parent-scan data to already imported scanned children even when those children are absent from the incremental REST issue response.

The default `KATA_GITHUB_TOKEN` is now scoped to `github.com`. Enterprise bindings require a matching App credential, `[github_sync].token_host = "<host>"` for the configured env token, or `gh auth token --hostname <host>` on the daemon host.

The branch intentionally does not add writable GitHub sync or broad sync-mode policy. Local kata parent links still win when they conflict with a GitHub-sourced parent insert.


wesm:
looking at this

wesm:
Sorry I kinda co-opted this PR to do some infrastructure improvements!

PR #150, opened 2026-07-03
Add branch-orchestration primitives design sketch

wesm:
## Summary

Design sketch (RFC — no implementation yet) for letting kata serve as the coordination substrate for branch-based agent orchestration, posted for comment before any implementation work.

The doc proposes a layered, metadata-first approach with a hard constraint that nothing adds complexity for existing kata users:

- **Layer 0 (small code):** expose the existing issue-metadata patch endpoint through the CLI (`kata meta set/unset/get`), accept `metadata` at issue creation, and add a `list --meta` filter.
- **Layer 1 (docs only):** documented `work.*` metadata conventions — `work.branch`, `work.attention` (`ok|attention|blocked`), `work.attention_msg` — plus a tracking-issue recipe covering launcher, working agent, coordinator, and merge automation roles, with explicit scope/concurrency/ownership semantics.
- **Layer 2 (deferred):** promote `work.*` into the reserved-key registry with validators, `kata wait`, and TUI attention badges, only once usage proves out.

The sketch documents what the daemon already provides (per-key merge patch with If-Match concurrency and per-key diff events, the reserved-key registry with opaque pass-through, federation claim gates on the metadata path) and names [middleman](https://github.com/kenn-io/middleman) as the reference consumer via its existing kata workspace-launch design. Orchestration itself — worktrees, terminals, lifecycle status — is explicitly out of scope for kata.

Four open questions for reviewers are listed at the end of the doc: key naming style, whether `list --meta` ships in Layer 0, the CLI shape (`kata meta` vs `edit --meta`), and whether `work.attention` should be registry-validated from day one.

### DIFF
--- a/docs/superpowers/specs/2026-07-03-branch-orchestration-primitives-design.md
+++ b/docs/superpowers/specs/2026-07-03-branch-orchestration-primitives-design.md
@@ -1,0 +1,190 @@
+# Branch-orchestration primitives for kata — exploratory sketch
+
+Status: **direction confirmed 2026-07-03 — coordination substrate, metadata-first.**
+This sketch records a feature survey of an adjacent per-branch agent
+orchestrator (reviewed 2026-07-03) and proposes which of its ideas kata should
+grow. Maintainer constraints, confirmed the same day:
+
+- Kata stays a tracker; orchestration (worktrees, terminals, launch) lives in
+  external consumers.
+- **Nothing may add complexity for existing kata users.** New capability
+  arrives as additional fields/metadata on issues and additive read paths,
+  not as new required concepts.
+
+The reference consumer is
+[middleman](https://github.com/kenn-io/middleman), which already ships a kata
+mode (talking to external kata daemons via `$KATA_HOME/config.toml`
+discovery) and a kata-task-workspace-launch design that maps a kata task to
+one tracked repository and creates a worktree keyed by the task's UID.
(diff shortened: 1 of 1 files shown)


mjacobs:
Took you up on "take a crack at this" — the implementation is up as #153.

It includes the design doc from this PR, amended in place; treat the amendments as comments on your RFC in diff form, all structured to be easy to revert piecemeal:

- the three adjustments I'd floated on Discord (hooks blessed as first-class `work.attention` writers alongside the agent; `kata wait` promoted to its own layer right after the conventions; attention enum renamed to `ok | needs-human | stuck` to avoid the `blocked` collision);
- proposed answers to your four open questions, marked in the doc as proposals rather than decisions (keep the dotted `work.` prefix; `list --meta` in Layer 0; a dedicated `kata meta` verb; `work.attention` stays registry-unvalidated for now).

One thing implementation surfaced, flagged prominently in #153 since it touches existing endpoint semantics: the metadata patch endpoints required `If-Match`, which forced every writer into read-modify-write with spurious 412s under concurrency — contradicting the doc's stated last-write-wins default for convention keys. #153 makes the header optional there (present = exact optimistic-concurrency behavior as before; move/recurrence endpoints unchanged). Isolated in its own commits if you'd rather keep it mandatory.

Leaving this PR open for you to land or close as you prefer — no rush on any of it.

PR #153, opened 2026-07-07
Add branch-orchestration primitives: metadata plumbing, work.* conventions, kata wait

mjacobs:
## Summary

Builds the branch-orchestration primitives sketched in #150 — Layer 0, Layer 1, and the `kata wait` fan-out/join primitive. Three adjustments I proposed in our Discord discussion are folded into the design doc in this branch (the doc commit from #150 is included here and amended in place). They were not individually signed off — each is listed below so it is easy to veto:

- **Hooks are first-class attention writers.** Pure agent self-assertion under-delivers (agents forget to clear or raise attention), so the Layer 1 contract and the operations recipe explicitly bless launcher-installed harness hooks — e.g. a session-stop/idle hook — as legitimate writers of `work.attention`, alongside the agent itself.
- **`kata wait` promoted to its own layer, shipped now.** It is the fan-out/join primitive that makes the conventions a substrate for agents rather than just dashboards, and it is additive + read-only, so it ships right after the Layer 1 conventions instead of waiting behind badges/TUI (which stay deferred in what is now Layer 3).
- **Attention enum renamed to `ok | needs-human | stuck`** so `blocked` does not collide with kata's blocks/blocked-by dependency semantics. `needs-human`: the agent wants human input/review (may still be progressing); `stuck`: it cannot proceed.

The doc's four open questions get proposed resolutions in a new "Proposed decisions" section — my picks at execution start, not ratified: keep the dotted `work.` prefix; `list --meta` ships in Layer 0; dedicated `kata meta` verb; `work.attention` stays registry-unvalidated (promotion still deferred; `kata wait` interprets the enum client-side).

## What ships

**Layer 0 — metadata plumbing (code)**
- `kata meta set <ref> <key> <value>` (`--json-value`, `--if-match`), `kata meta unset`, `kata meta get [key]` over the existing patch endpoint, honoring the json/agent output modes.
- `metadata` accepted at issue creation (validated through the same registry path: reserved keys via type validators, unknown keys opaque, JSON null rejected) plus `kata create --meta key=value`. The `issue.created` event payload now carries the real metadata, so federation folds and event followers see creation-time metadata.
- `kata list --meta key[=value]` (repeatable; presence / string-equality; AND semantics) via `json_each` bound-key predicates — flat dotted keys like `work.branch` are literal keys, never JSON paths. `kata show` renders non-empty metadata in the human and agent modes. OpenAPI artifacts regenerated.

**Layer 1 — conventions (docs)**
- `docs/reference/metadata.md`: the metadata model, reserved keys vs opaque pass-through, CLI usage, and the `work.*` contract (scope, last-write-wins concurrency, ownership including hooks, and the `stuck`-not-`blocked` disambiguation).
- `docs/operations/agent-orchestration.md`: the tracking-issue recipe — launcher / working agent / coordinator / merge automation roles, hook wiring as the recommended default, a `kata wait` fan-out/join worked example, close-with-evidence, and failure modes.

**Layer 2 — `kata wait`**
- `kata wait <ref>... [--until closed|attention|needs-human|stuck] [--timeout <dur>] [--any|--all] [--poll-interval <dur>]`: read-only state polling; a close also completes attention-mode waits (reason distinguishes); dedicated timeout exit code (8); pure condition evaluator kept separate so an event-driven fast path can replace the poll loop without touching semantics.

## ⚠️ Behavioral change surfaced during implementation (review me)

**`If-Match` is now optional on the metadata patch endpoints.** This is the one change to existing endpoint semantics in the PR, and it was not part of the sketch or our discussion. Implementation surfaced that the endpoints required the header, forcing every writer into read-modify-write with spurious 412s under concurrency — contradicting the design's stated last-write-wins default for convention keys. Absent header = unconditional patch; present header keeps exact optimistic-concurrency behavior; move/recurrence endpoints still require it. Isolated in its own commits and easy to revert if you'd rather keep the header mandatory.

Everything else is opt-in and invisible to existing users; no new required concepts, no schema columns.

Includes the design doc from #150, amended as described above — treating the amendments as comments on your RFC in diff form. Leaving #150 open for you to close or land as you prefer.


mjacobs:
Resolved both Medium agent-mode output-contract findings from the roborev review in `a262350`.

- **`cmd/kata/wait.go` — abandoned ref emitting `ERR` on stdout**: in `--any` mode an abandoned ref now renders as a non-error `OK wait <ref> reason=abandoned` row instead of `ERR wait ... reason=error`. `ERR` is reserved for the top-level all-abandoned failure (stderr, nonzero exit), so a wait that succeeds on another ref no longer emits a spurious error line.
- **`cmd/kata/meta.go` — `meta get --agent` value quoting**: `writeMetaAgentRow` now routes both key and value through `writeAgentKVRow`/`agentRowField`, so a JSON value containing spaces, quotes, or backslashes stays a single agent-quoted token. This matches how `show --agent` already renders metadata rows.

Added a regression test for each (`TestWaitAnyAbandonedAgentUsesNonErrorRow`, `TestMetaGetAgentQuotesValueWithSpaces`); the `cmd/kata` package passes.

mjacobs:
Resolved the Medium finding from the review of `a262350` in `bfbae11`.

- **`cmd/kata/wait.go` — negative `--timeout` blocks forever**: a negative duration passed the poll loop's `timeout > 0` deadline gate and was silently treated like `0` (wait forever). `runWait` now rejects `opts.timeout < 0` during option validation (alongside the existing `--poll-interval` check) with an `ExitValidation` error, so a typo like `--timeout=-1s` fails fast. Added a `TestWaitUsageNegativeTimeout` regression test.

mjacobs:
Resolved the Medium finding from the review of `bfbae11` in `4562ff0`.

- **`cmd/kata/wait.go` — `--timeout` didn't bound the HTTP fetches**: the poll loop scheduled against a deadline but the state GETs ran on the unbounded command context, so a hung daemon request could block past `--timeout`; the loop could also start one more fetch pass after sleeping to the deadline. When `--timeout > 0`, all wait fetches (initial pass + poll loop) now run on a deadline-scoped context so a stuck request is cancelled at the deadline, while Ctrl-C cancellation still flows through the parent context used for the inter-poll sleep. Added a post-sleep deadline re-check so no fetch pass fires once the timeout has elapsed.

Added `TestWaitTimeoutBoundsHungFetch`, a stub-server test that hangs the issue GET and asserts `wait` returns `ExitWaitTimeout` bounded by `--timeout` (~0.3s) instead of blocking on the stuck request (verified it blocks the full ~5s without the fix).

mjacobs:
Resolved the Medium finding from the review of `4562ff0` in `2733f6d`.

- **`cmd/kata/wait.go` — `pending` populated on success**: the `--json` payload documents `pending` as timeout-only (refs still unmet when the wait times out), but it was filled on every run. On a successful `--any` join the later refs are intentionally never satisfied, so they appeared in `pending` and made a successful result read as incomplete. `Pending` is now populated only when `timed_out` is true; a successful join reports an empty pending list (the timeout path is unchanged and still lists the unmet refs).

Added `TestWaitAnyJSONSuccessLeavesPendingEmpty`, a `--json --any` success test with a second ref that never satisfies, asserting `pending` stays empty.

mjacobs:
Caught one more locally (via roborev on this branch) before CI and fixed it in `21c57e5`.

- **`cmd/kata/wait.go` — deadline-canceled fetch could surface as `ExitInternal`**: the `--timeout` deadline context added in `4562ff0` counted a deadline-canceled fetch toward the consecutive-failure budget (`maxWaitConsecutiveFails`). If a ref already had two prior transient (5xx) failures, the third deadline-canceled request tripped the budget and returned `ExitInternal` instead of `ExitWaitTimeout`. The poll loop now returns the normal timeout result when a fetch errors after the deadline has already elapsed, rather than counting it as a failure.

Added `TestWaitTimeoutAfterTransientFailsReportsTimeout`: a stub server primes two transient 5xx failures then hangs the next fetch past the deadline, asserting `ExitWaitTimeout` (red before the fix returned `ExitInternal`).

mjacobs:
Ran a whole-branch roborev review locally (the same `default_security` correctness+security panel CI uses) and cleared both findings it surfaced — both before CI.

- **Medium — `cmd/kata/wait.go`: `--any` initial pass could hang** (`da8719b`): in `--any` mode the initial pass kept fetching the remaining refs even after one ref already satisfied the join. If a later ref's fetch stalled and no `--timeout` was set, `kata wait closed-ref other-ref --any` could hang despite the wait already being met. The initial pass now breaks as soon as a ref satisfies the wait in `--any` mode (matching the poll loop). Added `TestWaitAnyInitialPassStopsAfterJoinMet` (stub server: first ref closed, second ref stalls → returns promptly).

- **Low — `internal/daemon/handlers_metadata.go`: present-but-empty `If-Match` bypass** (`7d01a0d`): the metadata patch endpoints treated a present-but-empty `If-Match` header the same as an absent one, silently downgrading a malformed conditional write to unconditional last-write-wins (huma's typed binding reads headers via `Header.Get`, collapsing absent and present-empty to `""`). Added a resolver on both patch request types that inspects the raw headers via `huma.Context.EachHeader` and rejects a present-but-empty `If-Match` with the standard 400; an absent header stays the documented unconditional default. Added `TestPatchMetadata_PresentEmptyIfMatch_400` for both the issue and project subjects.

The security member of the panel passed.

wesm:
Matt, I could not push directly to your fork branch, so I pushed the follow-up fixes to a branch I can write:

https://github.com/kenn-io/kata/tree/wesm/branch-orchestration-fixes

The two commits to pick up are:

- 506896a4 Make wait timeout cover ref resolution
- 9cf8d0bf Satisfy wait timeout test lint

The push hooks now pass locally: `make api-check`, `make lint`, and `nilaway`. I also ran `go test ./...` and `docs/zensical-docs.sh build` after the timeout/docs changes.

mjacobs:
Follow-up now that this is merged and in daily use: **the design doc's acceptance loop has run end-to-end on real work, twice** — coordinator sessions driving ready kata tasks in [mjacobs/tabby](https://github.com/mjacobs/tabby) and [mjacobs/yas](https://github.com/mjacobs/yas), fanning out to sub-agents in git worktrees. Small projects, but live boards and real shipped commits — not a contrived harness.

**Setup, for reproducibility** — two pieces of wiring, both explicit for now:

1. A short `work.*` conventions section in each repo's `AGENTS.md` (claim → set `work.attention ok`; stamp `work.branch`; raise `stuck`/`needs-human` with an `attention_msg`; close or hand off before ending). Happy to PR this into `agent-orchestration.md` as an "adopting in your repo" copy-paste section.
2. An explicit coordinator prompt: *"pick two independent tasks and run them in parallel: claim each with `--meta work.branch=...`, delegate to a subagent in its own worktree, and join with `kata wait <refs> --until attention --any`."*

What the event log shows:

- **Fan-out:** two sub-tasks per run, claimed and branch-stamped — one via `--meta` at create, one via `kata meta set` after; both write paths exercised.
- **Join:** the coordinator blocked on `kata wait <a> <b> --until attention --any --timeout 30m` while the workers ran.
- **Attention mode:** a worker raised `needs-human` ("implemented, tests green, ready for review/merge") → the wait fired → the coordinator reviewed, merged, closed with commit evidence, and cleared the attention pair. Full cycle, no human in the loop.
- **Close mode:** in the yas run the wait terminated on an issue *close* instead, and the reported reason distinguished the two correctly — both termination modes have real reps.
- **The hand-off state:** one tabby branch ended "merged, but a browser-only check remains" — left open as `needs-human` with an `attention_msg` saying exactly what's pending. `kata list --meta work.attention` then showed exactly one row, and it was correct. The "which branch needs a human" board works.

One adoption finding worth feeding back into the recipe's hooks paragraph: **compliance tracked instruction explicitness monotonically.** Same board, same CLI — no conventions in context: `work.*` written on 0/2 tasks; conventions in `AGENTS.md` only: 1/3, inconsistently within a single session; explicit prompt: 2/2 with the full lifecycle. Documentation produces tendency, not contract — the launcher-installed-hooks story really is the reliability floor, not an optimization. We're experimenting on our side with making that wiring disappear (extending the `--with-agents` block, and launcher-installed hooks per the recipe) and will bring issues/PRs once it's proven in use rather than proposals now.

One papercut, filed as #159: the first real agent invocation of `kata wait` used `--timeout 1800` and got rejected for the missing unit — agents naturally write bare seconds.

PR #161, opened 2026-07-09
Suggest the unit when a duration flag gets a bare number

### LINKED ISSUE #159: kata wait: accept unitless --timeout as seconds (or hint the unit in the error)
Observed in real agent use (coordinator session driving the fan-out/join loop from `docs/operations/agent-orchestration.md`): the first `kata wait` invocation an agent ever composed was

```
kata wait 6yae qdxf --until attention --any --timeout 1800 --agent
```

which failed on duration parsing — the agent assumed bare seconds, then self-recovered with `--timeout 30m` (~25s and one turn lost). Agents (and plenty of humans) write unitless timeouts by default, and `wait` is the primary agent-facing verb, so this is a first-contact papercut worth smoothing.

Either fix works:

1. Accept a bare integer as seconds for `--timeout` / `--poll-interval`.
2. Keep strict parsing but make the error actionable: `invalid duration "1800": missing unit — did you mean "1800s"?`

Happy to send a PR for whichever direction you prefer.

mjacobs:
## Summary

Fixes #159 (option 2, per the precedent survey there): duration flags keep the units-required grammar, but a bare number now fails with a suggestion of the seconds spelling instead of Go's stock parse error, and non-numeric input gets example spellings.

```
$ kata wait abc4 --timeout 1800
kata: invalid argument "1800" for "--timeout" flag: missing unit in duration "1800": did you mean "1800s"? (units: s, m, h)

$ kata wait abc4 --timeout soon
kata: invalid argument "soon" for "--timeout" flag: invalid duration "soon": use a Go duration such as 30s, 5m, or 1h30m
```

## What ships

- `durationFlag`, a small `pflag.Value` wrapping `time.ParseDuration` with the friendlier failure modes (`cmd/kata/duration_flag.go`). Follows the existing "strict grammar, generous errors" precedents (`digest`'s `parseSinceUntil`, `internal/hooks`' `parseDuration`).
- Wired into `kata wait --timeout` and `--poll-interval` (previously stock `DurationVar`). Accepted values are unchanged — `0` still parses (Go accepts unit-less zero), negative durations still reach the command's existing range validation.
- Tests: unit coverage for the suggestion/guidance/valid/default-rendering paths, plus a CLI-level regression for the exact real-world failure (`--timeout 1800`).

## Deliberately not included

`--ttl` (federation lease) and `--interval` (github-sync) are `StringVar` today and parsed later; converting them to `durationFlag` also moves their validation client-side, which is a behavior question worth its own decision. The type is reusable if you want them switched — happy to follow up either way.

### DIFF
--- a/cmd/kata/duration_flag.go
+++ b/cmd/kata/duration_flag.go
@@ -1,0 +1,41 @@
+package main
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+)
+
+// durationFlag is a pflag.Value for duration flags that keeps Go's
+// units-required grammar but fails with guidance instead of the stock
+// parser error. The common mistake — agents especially — is a bare number
+// ("--timeout 1800"); bare numbers stay rejected because they are ambiguous
+// (seconds? milliseconds?), but the error suggests the seconds spelling
+// instead of just naming the parse failure.
+type durationFlag struct {
+	d *time.Duration
+}
+
+func (f durationFlag) String() string {
+	if f.d == nil {
+		return "0s"
+	}
+	return f.d.String()
+}
+
+func (f durationFlag) Set(s string) error {
+	trimmed := strings.TrimSpace(s)
+	d, err := time.ParseDuration(trimmed)
+	if err != nil {
+		if _, numErr := strconv.ParseFloat(trimmed, 64); numErr == nil {
+			return fmt.Errorf("missing unit in duration %q: did you mean %q? (units: s, m, h)",
+				trimmed, trimmed+"s")
+		}
+		return fmt.Errorf("invalid duration %q: use a Go duration such as 30s, 5m, or 1h30m", trimmed)
+	}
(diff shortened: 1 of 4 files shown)


mariusvniekerk (approved):

mariusvniekerk:
Why not just use the actual duration flag from cobra ?

mjacobs:
> Why not just use the actual duration flag from cobra ?

Good question— it was `DurationVar` before this PR (see the removed lines in `wait.go`). Cobra/pflag's built-in duration flag surfaces Go's stock `time: missing unit in duration "1800"` error, which was the friction I found in my testing -> #159 . `pflag` doesn't let you customize a built-in type's `error`, so this wraps the same `time.ParseDuration` in a `pflag.Value` purely to make the failure suggest the fix (`did you mean "1800s"`?). It is extra code but the parsing grammar is unchanged, and easy to remove if there's an alternative later.

PR #164, opened 2026-07-10
Converge federated cross-project links

mariusvniekerk:
## Why

Federation currently quarantines a valid batch whenever a link peer has not reached the hub yet. If two independently synchronized projects create cross-references in their first pending batches, each waits for an issue that only the other quarantined batch can create, permanently blocking both push cursors.

## What changes

- accept valid link events with unresolved peers while preserving strict validation for primary issues, actor violations, malformed events, and hash conflicts
- reconcile links across enabled hub projects on one daemon or spoke projects sharing the same normalized hub origin, so arrival order no longer matters
- keep unfederated and different-origin peers absent without blocking subsequent events
- ensure CI test commands cannot inherit KATA_AUTH_TOKEN
- document same-hub enrollment constraints and retry-based recovery for quarantines created by older builds

Missing peers remain event-backed deferred state; no schema migration or separate pending-link store is added. Existing quarantines should be retried, not skipped, after compatible hub and spoke binaries are deployed.

The full shuffled Go suite, lint, API compatibility check, workflow validation, documentation checks, and push hooks passed on the committed branch.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,7 +27,7 @@ jobs:
           cache: true
 
       - name: Run tests
-        run: go test -p 1 ./...
+        run: env -u KATA_AUTH_TOKEN go test -p 1 ./...
 
   windows:
     name: Windows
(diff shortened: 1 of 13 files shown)


wesm:
looking

PR #172, opened 2026-07-10
Add kata init --with-hooks: install work.attention hooks for Claude Code

mjacobs:
Implements the installer proposed in #162: `kata init --with-hooks` writes the attention-hook wiring from `docs/operations/agent-orchestration.md` ("Keep attention truthful with hooks") into a Claude Code workspace, so the reliability floor no longer depends on hand-wiring.

What it installs:

- A managed script at `.claude/hooks/kata-attention.sh` (embedded in the binary, refreshed on drift). Three modes:
  - `start` (SessionStart) — honors `$KATA_REF` from a launcher and floors that issue to `work.attention ok`;
  - `claim` (PostToolUse, Bash matcher) — answers the ref-discovery question for hand-started sessions by treating the agent's own `kata claim <ref>` as the discovery event, recording refs in a per-session state file;
  - `stop` (Stop) — sweeps recorded refs still sitting at `ok` up to `needs-human` / "session ended without hand-off".
- Additive `.claude/settings.json` wiring for those three events. The merge preserves everything else in the file (unknown keys, existing hooks), is idempotent across re-runs, reports rather than clobbers a malformed file, and refuses symlinked targets — matching the `--with-agents` posture. `--with-hooks` and `--with-agents` compose in one init.

On the open design questions in #162: ref discovery is start-env + claim-watch as above; the surface is the sibling-flag option (`kata init --with-hooks`), idempotent alongside the managed guidance block; harness scope is Claude Code first, with the script body kept to the generic shell shape so other harnesses can reuse it.

Beyond unit/e2e coverage, the script was exercised against a live daemon on a seeded scratch board: launcher start, claim discovery, non-claim commands ignored, stop sweep raising `needs-human`, closed/unknown refs skipped end-to-end (that dry-run caught — and this PR guards — the claim-time floor writing onto a closed issue), mid-session `stuck` self-assertion surviving the sweep, and repeat-stop idempotency.

---

**Gated on #174.** Review found the repo-committed hook script is a code-execution vector (approval covers the settings.json command string, not the script content it points at). #174 reworks this installer to embed the hook logic in the kata binary; this PR is draft until that lands or the maintainer prefers them folded together.

PR #174, opened 2026-07-11
Move the attention hook into the kata binary

mjacobs:
Supersedes the script half of #172 (whose commits this branch includes — the diff on top is the rework), responding to the review finding that a repo-committed hook script is a code-execution vector: Claude Code re-prompts when a settings.json hook *command* changes, but the command string `.claude/hooks/kata-attention.sh` stays constant while a hostile commit can rewrite the script's content after first approval.

The hook logic now lives in the installed binary as a hidden plumbing command, and `kata init --with-hooks` wires `.claude/settings.json` to call it directly:

- SessionStart → `kata attention-hook start`
- PostToolUse (Bash) → `kata attention-hook claim`
- SessionEnd → `kata attention-hook stop`

What executes is now exactly what the approval dialog shows, and no executable content is written into the repository. It also retires the whole class of problems inherent to a shell artifact — execute-bit management, BSD/bash-3.2 portability, the jq dependency — and made the two remaining review findings on the script fixable properly:

- Claim parsing is a real two-pass tokenizer: global flags on either side of `claim` (`kata --project spoke-project claim abc4`), value-bearing flags with space or `=` forms, quoted arguments with escape handling, and one effective actor (`--as` never falls back to ambient) resolved before any probing.
- Routing context is honored end-to-end: a claim's `--project`/`--workspace`/`--daemon` route its verification and mutations, and session state stores project-qualified refs so the SessionEnd sweep resolves the same issue the claim did. (Known limitation, documented inline: a claim against a different daemon/workspace can't be swept from the ambient session context.)
- Upgrading is destructive toward the old layout on purpose: refreshing prunes every legacy script-based hook entry across all events, so no previously-approved script command survives pointing at stale on-disk content.
- Session state lives under `XDG_RUNTIME_DIR` (or the user cache dir — never world-writable /tmp), with session-id validation so hook input can't traverse outside it.

Hook semantics are unchanged from #172: launcher-supplied `$KATA_REF` at start, `kata claim` discovery for hand-started sessions gated on daemon-verified ownership, and the terminal ok→needs-human sweep with transient-failure retry. Behavior that was previously verified by hand is now pinned by parser unit tests and live-daemon e2e tests fed real hook JSON over stdin.

### DIFF
--- a/cmd/kata/attention_hook.go
+++ b/cmd/kata/attention_hook.go
@@ -23,6 +23,7 @@ const (
 	attnValueOK         = "ok"
 	attnValueNeedsHuman = "needs-human"
 	attnHandoffMsg      = "session ended without hand-off"
+	attnWriteAttempts   = 2
 )
 
 func newAttentionHookCmd() *cobra.Command {
@@ -82,8 +83,16 @@ type attnLookup struct {
 
 type attnDaemon interface {
 	lookup(ref string) attnLookup
-	setMetaIfRevision(ref string, patch map[string]string, revision int64) bool
+	setMetaIfRevision(ref string, patch map[string]string, revision int64) attnWriteResult
 }
+
+type attnWriteResult uint8
+
+const (
+	attnWriteFailed attnWriteResult = iota
+	attnWriteApplied
+	attnWriteConflict
+)
 
 func attentionRef(kataRef string) (string, bool) {
 	ref := strings.TrimSpace(kataRef)
@@ -97,9 +106,14 @@ func attnStart(d attnDaemon, kataRef string) {
 	if !ok {
 		return
 	}
-	lookup := d.lookup(ref)
-	if lookup.kind == lookupOpen {
-		d.setMetaIfRevision(ref, map[string]string{attentionKey: attnValueOK}, lookup.revision)
+	for range attnWriteAttempts {
+		lookup := d.lookup(ref)
+		if lookup.kind != lookupOpen {
+			return
+		}
+		if d.setMetaIfRevision(ref, map[string]string{attentionKey: attnValueOK}, lookup.revision) != attnWriteConflict {
+			return
+		}
 	}
 }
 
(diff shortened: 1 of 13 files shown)


mjacobs:
### Design note: the claim hook is intentionally execution-ambiguous

Flagging this for reviewers (human and automated) so a recurring finding is understood as a deliberate trade-off rather than a bug.

**Context.** The `claim` mode of the attention hook runs as a Claude Code `PostToolUse` hook on `Bash`. It parses the tool's command string for a `kata claim <ref>` invocation. That JSON payload carries **no exit code**, so the hook cannot know whether the parsed command actually executed or succeeded.

**Why we parse `&&`-chained forms anyway.** An earlier review round established that `cd /some/dir && kata claim <ref>` is the dominant shape agents use, so command-position detection deliberately matches a `kata claim` after `&&`. The consequence is unavoidable: `false && kata claim <ref>` (a guard whose left side failed, so the claim never ran) is textually indistinguishable from the legitimate form. Guarding `&&` at the parser would regress the dominant real case, so we don't.

**The accepted residual.** Because execution can't be confirmed from text, a possibly-unexecuted claim can, in narrow cases, produce a **spurious or missed `needs-human`** — but only on an issue the **acting identity already owns** (the daemon-ownership oracle gates every action). We accept this bounded residual rather than over-engineering shell/exit-code emulation into the hook.

**What is strictly enforced** (so the residual stays bounded and safe):

- A claim **never downgrades a deliberate non-`ok` attention value** (`needs-human`/`stuck`) back to `ok` — the floor is gated on the daemon's current value.
- A claim on an issue already at a deliberate non-`ok` value **records nothing** — it is already a human-visible signal, and registering an unconfirmed claim would risk a spurious escalation if it later returned to `ok`.
- The terminal (SessionEnd) sweep **re-validates every ref against the daemon** before acting, and only escalates issues still sitting at the `ok` baseline.
- The floor writes `ok` **only after** the ref is durably recorded for the sweep, so no issue is floored without sweep coverage.

**Net:** an automated reviewer may flag "the hook acts on a claim it can't prove executed" as medium severity. That is this documented, intentional trade-off. The harm is bounded to a self-owned issue's attention flag; the correctness-critical property — never clobbering someone else's issue, and never silently downgrading a deliberate hand-off — is preserved.

wesm:
looking

wesm:
looking

Issue #162, opened 2026-07-09
Tooling to install the work.attention harness hooks from the orchestration recipe


mjacobs:
`docs/operations/agent-orchestration.md` documents launcher-installed hooks as the recommended default for keeping `work.attention` truthful — a session-start hook setting `ok` and a stop/idle hook raising `needs-human` if the agent never handed off. Today that wiring is entirely manual: the doc shows the shell snippets, but nothing installs them.

Field evidence that this matters (details in the [dogfood report on #153](https://github.com/kenn-io/kata/pull/153#issuecomment-4928446663)): compliance tracked instruction explicitness monotonically — with no conventions in context, agents wrote `work.*` on 0/2 tasks; with the conventions in `AGENTS.md` only, 1/3 *within a single session*; with an explicit prompt, 2/2 with the full lifecycle. Docs and context hints produce tendency, not contract; hooks are the only writer whose reliability doesn't depend on the agent attending to instructions.

Proposal: a command that installs the recipe's two hooks into a harness config — e.g. under the same surface as `kata init --with-agents` (which already manages an agent-guidance block), or a sibling like `kata init --with-hooks`. Claude Code's `.claude/settings.json` would be the first target since its hook events map directly onto the recipe; the hook bodies are the generic shell commands already in the doc, so other harnesses that can run a command at session start/stop follow the same shape.

Open design questions:

1. **Ref discovery.** The recipe's hooks take `$KATA_REF` from the launcher's environment — correct for orchestrator-launched sessions, but a hand-started session has no launcher to pass it. Options: resolve the tracking issue from the current branch via `work.branch` (a `kata list --meta work.branch=$(git branch --show-current)` lookup), read a workspace-local file, or simply no-op when unset (hooks only fire for launched sessions).
2. **Command surface** — extend `--with-agents`, a new flag, or a separate verb; and how it stays idempotent/refreshable alongside the managed `AGENTS.md` block.
3. **Which harness configs** are in scope for a first cut vs. documented-only.

We're prototyping this locally (hand-wired `.claude/settings.json` in the repos from the dogfood run) and will attach results here — if the gradient logic holds, hooks should turn the 1/3 arm into 3/3 with no prompt nudging, which is the result that justifies the tooling.

mjacobs:
[...]

PR #191, opened 2026-07-21
Add kata init --with-codex-hooks: Codex CLI attention-hook target

mjacobs:
Extends `kata init` so Codex CLI workspaces can participate in kata's
`work.attention` lifecycle. The new `--with-codex-hooks` flag installs the
Codex start hook while remaining independent from `--with-agents` and the
existing Claude Code `--with-hooks` option.

## Behavior

- Adds a `SessionStart` command hook to `.codex/hooks.json` that runs
  `kata attention-hook start`.
- Limits the hook to startup, resume, and clear transitions. Context compaction
  does not reset a live `needs-human` or `stuck` signal.
- Uses the launcher's `KATA_REF` environment variable, matching the existing
  Claude Code attention workflow.
- Migrates matcher-less kata hooks written by earlier versions of this branch
  to the scoped matcher on the next `kata init --with-codex-hooks` run.

## Session end

Stable Codex releases do not yet provide a session-end hook suitable for this
workflow, so this flag installs only the native start half. The agent
orchestration documentation includes a launcher-wrapper recipe that runs
`kata attention-hook end` when Codex exits; native end-hook wiring remains a
follow-up once the upstream event is stable.

## Configuration safety

The installer merges into the existing hooks file instead of replacing it. It
preserves unrelated events and metadata, is byte-stable on rerun, refuses
malformed files and symlinked targets, and atomically replaces an existing
file. If `.codex/config.toml` already defines hooks, kata emits a non-fatal
warning because Codex loads both sources.

The CLI reference, agent workflow, orchestration guide, and changelog describe
the new flag and the native-start/launcher-end workflow.

Extends #174 and advances #162.

### DIFF
--- a/cmd/kata/init.go
+++ b/cmd/kata/init.go
@@ -19,20 +19,22 @@ import (
 
 // initOptions holds the flags specific to `kata init`.
 type initOptions struct {
-	Project    string
-	Replace    bool
-	Reassign   bool
-	WithAgents bool
-	WithHooks  bool
+	Project        string
+	Replace        bool
+	Reassign       bool
+	WithAgents     bool
+	WithHooks      bool
+	WithCodexHooks bool
 }
 
 // callInitOpts is the parameter bag passed to callInit.
 type callInitOpts struct {
-	Project    string
-	Replace    bool
-	Reassign   bool
-	WithAgents bool
-	WithHooks  bool
+	Project        string
+	Replace        bool
+	Reassign       bool
+	WithAgents     bool
+	WithHooks      bool
+	WithCodexHooks bool
 }
 
 // cliError is a structured error that carries an exit code for main().
@@ -142,6 +144,7 @@ get committed.`,
 	cmd.Flags().BoolVar(&opts.Reassign, "reassign", false, "move an existing alias to this project")
 	cmd.Flags().BoolVar(&opts.WithAgents, "with-agents", false, "write kata agent guidance into AGENTS.md/CLAUDE.md in the workspace")
 	cmd.Flags().BoolVar(&opts.WithHooks, "with-hooks", false, "install work.attention harness hooks into the workspace's Claude Code config (.claude/)")
+	cmd.Flags().BoolVar(&opts.WithCodexHooks, "with-codex-hooks", false, "install Codex CLI work.attention hook wiring (.codex/hooks.json)")
 
 	return cmd
 }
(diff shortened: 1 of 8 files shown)


mjacobs:
addressing the finding above

mjacobs:
Re the High in the `ceef859` review above ("Repository-controlled executable can impersonate `kata`", `cmd/kata/init_hooks_codex.go:33`): triaged as an **accepted trust boundary**, not fixed — the bare `kata` command is deliberate.

- Pinning an absolute path at install time (`os.Executable()`) would make `.codex/hooks.json` machine-specific. The file is workspace-level and may be committed or shared, so a pinned path breaks the hook for every other machine or user of the workspace.
- The shipped Claude Code wiring makes the identical choice (bare `"kata"` command in `.claude/settings.json`), so this PR adds no new surface relative to what is already on `main`.
- A `PATH` that resolves `kata` to a repository-controlled binary means the developer's shell is already compromised — every `make`/`npm`/git-alias invocation has the same exposure — which puts a hostile `PATH` outside this installer's threat model.

The boundary is now documented inline at the flagged site (4e00944) so future reviews can triage against it.

mariusvniekerk:
Human Marius here, So what does this actually get us?

mjacobs:
> Human Marius here, So what does this actually get us?

Yeah, good question: I'm trying to close the gap between "the agent *should* maintain the attention signal" and "the signal is maintained"-  this is the Codex port of the installer (#162).

**Today, without hooks**, driving a Codex session against a tracked issue means spending prompt space on bookkeeping and *hoping* it's honored:

```sh
export KATA_REF=abc4
codex "Work abc4. First set work.attention=ok; if you stop for any reason,
       close the issue or set needs-human with an attention_msg."
```

When the agent forgets the exit step (they do, in my local experimentation), the issue sits at a stale `ok`: the board says "healthy, being worked" while nobody is working it.

**With this PR**, `kata init --with-codex-hooks` can be issued once per workspace, then you can drive a Codex session like:

```sh
export KATA_REF=abc4
codex "Work abc4."
```


That said, I'm still hoping to make this story easier...

I've been thinking about these ideas below:

1. **(A)**  Native session-end (for Codex): The wrapper exists only because Codex has no stable session-end hook event yet (openai/codex#20374; the `SessionEnd` event merged upstream in openai/codex#33895 but is alpha-only so far). Once in a stable release, `--with-codex-hooks` can expand to include `SessionEnd → kata attention-hook end` handler and the wrapper becomes nice-to-have instead of a requirement.
**(B)** Make the ref itself implicit: the launcher already exports `KATA_REF` so the adoption snippet in `agent-orchestration.md` could add one line like "if `KATA_REF` is set, that's your tracked issue". Then launched sessions are more succinct, like `codex "continue the tracked work"`.

3. **Covering hand-started sessions.** Today the hooks only bind to a session when a launcher exports `KATA_REF`. An earlier iteration (#172, reworked in #174) tried discovering the tracked issue by watching for `kata claim <ref>` in the agent's shell commands. It sometimes worked... but reconstructing "did a claim execute, and against what" from command text proved structurally leaky, so it was cut before merge. Better would be a daemon-side solution: the daemon already processes every claim with canonical identity, so the EOS sweep could query "what did this actor claim during the session window"- no parsing. Needs session correlation in the daemon which is more meaningful design work.


LMK if you have ideas/feedback

wesm:
looking at this

PR #195, opened 2026-07-21
Honor repeated query filter values

### LINKED ISSUE #189: When giving --label or --no-label multiple times, only the first instance is respected
I ran into an unexpected issue using `kata ready` with multiple `--no-label` flags: `kata ready --no-label a --no-label b`.

From the inline help (`kata ready --help`), I was expecting this to show me all unblocked kata tasks _excluding_ any tasks labelled with `a` or `b`. Instead, it excluded tasks labelled `a`, but failed to exclude tasks labelled `b`. It looks as though the `--no-label b` had no effect. (Here `a` and `b` are of course placeholders for real labels - the actual call looked like `kata ready --no-label icebox --no-label specification`.)

I also tried `kata ready --no-label a,b`, with the same result: only tasks with label `a` were excluded.

It looks as though the actual issue is broader, affecting commands like `kata list` and `kata next` in addition to `kata ready`, and applies to flags `--label` and `--meta` in addition to `--no-label`.

Seen on kata 0.11.1 (go1.26.3, linux/arm64), but also observed on latest main (commit 0abb14f783ed25161714f552006120c16d3f73a4).

I had Claude Code (Opus 4.8 / xhigh) investigate. Claude's (characteristically loquacious) write up, including instructions to reproduce and tentative diagnosis, is below the line. I haven't attempted to validate the diagnosis.

--- 

# `--no-label` / `--label` / `--meta` only honor the first value (repeated flag is silently dropped)

## Summary

Multi-value list filters that the help advertises as "repeatable" only ever
apply their **first** value. Passing a filter flag twice — or as a
comma-separated list — silently drops every value after the first. This
affects `kata list` and `kata ready`/`kata next` for `--label`, `--no-label`,
and (on `list`) `--meta`.

Single-value use works correctly; the bug only shows up once you try to
combine two or more values.

## Environment

- Reproduced on `main` (`kata g0abb14f`) and present unchanged in the tagged
  `v0.11.1` release (`09a6743`). Not a recent regression — the affected
  encoding dates back to PR #49 (2026-05-25).
- Backend: default local daemon (SQLite).

## Repro

```
kata init --project repro-proj
kata create "task alpha"      --label a
kata create "task beta"       --label b
kata create "task gamma"      --label c
kata create "task alpha-beta" --label a --label b
kata create "task beta-c"     --label b --label c
```

Now exclude both `a` and `b` — expected result is only the issue labelled
solely `c` ("task gamma"):

```
$ kata --agent list --no-label a --no-label b
OK list count=3
- issue=... labels=b,c title="task beta-c"     # <-- still has b, should be gone
- issue=... labels=c   title="task gamma"
- issue=... labels=b   title="task beta"       # <-- still has b, should be gone
```

Only `a` was excluded; the second `--no-label b` had no effect. The
comma-separated form behaves identically (also broken):

```
$ kata --agent list --no-label a,b        # same 3 rows — only 'a' excluded
```

Order-dependent, confirming "first value wins":

```
$ kata --agent list --no-label b --no-label a   # now only 'b' is excluded
```

`--label` (documented as AND logic) has the same defect — this returns issues
that have only `a`, not just those having both `a` and `b`:

```
$ kata --agent list --label a --label b
OK list count=2
- issue=... labels=a,b title="task alpha-beta"
- issue=... labels=a   title="task alpha"      # <-- lacks b, should be excluded
```

And multiple `--meta` filters on `list` (also documented AND):

```
kata create "meta both"     --meta team=core --meta tier=1
kata create "meta teamonly" --meta team=core
$ kata --agent list --meta team=core --meta tier=1   # returns both; only team=core applied
```

`kata ready --no-label a --no-label b` and `kata next` show the same behavior
(they share the same client query-building path).

## Root cause

The CLI encodes each repeated value as a **separate** query parameter, e.g.
`?exclude_label=a&exclude_label=b` (the `--label`/`--no-label` flags are
`StringSlice`, so `a,b` is split client-side and re-emitted as repeated
params — which is why the comma form and the repeated-flag form hit the
daemon identically).

The daemon's request structs decode these fields as comma-delimited,
single-occurrence query arrays (`[]string` with a `query:"..."` tag under
Huma v2). It reads only the **first** occurrence of the parameter and would
split *that* on commas; subsequent occurrences are ignored.

Probing the daemon HTTP API directly makes the mismatch explicit:

```
# repeated params (what the CLI actually sends) -> only 'a' excluded
GET /api/v1/projects/<id>/issues?exclude_label=a&exclude_label=b
  -> returns issues still labelled b

# single comma-joined param -> both excluded (correct)
GET /api/v1/projects/<id>/issues?exclude_label=a,b
  -> returns only the c-labelled issue
```

So the daemon's *comma-joined* form works correctly; the CLI never sends it.

## Affected surface

- `kata list --label` (AND)
- `kata list --no-label` (exclude)
- `kata list --meta` (AND)
- `kata ready --label` / `--no-label`
- `kata next --label` / `--no-label`

## Workaround

There is no working CLI invocation for combining values today: the
`StringSlice` flags split on commas and re-emit repeated params regardless of
quoting. The only current workaround is to call the daemon HTTP API directly
with a single comma-joined parameter (`?exclude_label=a,b`).

## Documentation angle

`kata list --help` renders these as `--no-label strings ... (repeatable)` and
`--label strings ... (repeatable, AND logic)`. The `strings` type and
`(repeatable)` note are accurate about the *flag parsing* — the CLI genuinely
accepts multiple values — but the end-to-end behavior does not honor them, so
the advertised capability is effectively broken.

## Suggested fix direction

Two options (not prescriptive):

1. **Daemon-side (most robust):** decode these query arrays so repeated
   occurrences are all read (accept `?label=a&label=b`). This keeps the
   existing CLI encoding and is the only option that is safe for `--meta`,
   whose values can legitimately contain commas.
2. **CLI-side:** send a single comma-joined param
   (`params.Set("label", strings.Join(labels, ","))`). Safe for
   `label`/`exclude_label` because the label charset (`[a-z0-9._:-]`) excludes
   commas, but **not** safe for `--meta` values, so `--meta` would still need
   the daemon-side fix.

Whichever path is chosen, a regression test that exercises two distinct values
through the full CLI→daemon→DB path (not just the DB query, which already
handles multiple values correctly) would lock this down.

wesm:
Repeatable list and ready filters were accepted and emitted as repeated query parameters, but daemon decoding discarded every occurrence after the first. That made the documented AND and exclusion semantics unreliable, including through `next`, and affected repeatable digest actor selection built on the same contract.

Opt the repeated array parameters into Huma exploded query decoding and add end-to-end coverage for label, exclusion, metadata, and actor filters. This preserves values verbatim instead of joining them, which matters for metadata filters that may contain commas.

Closes #189

### DIFF
--- a/cmd/kata/list_test.go
+++ b/cmd/kata/list_test.go
@@ -44,6 +44,33 @@ func TestList_DefaultsToOpenIssuesInProject(t *testing.T) {
 	assert.Contains(t, out, "beta")
 }
 
+func TestList_RepeatedLabelFiltersRequireEveryLabel(t *testing.T) {
+	env, dir, pid := setupCLIWorkspace(t)
+	alphaOnly := createIssue(t, env, pid, "alpha only")
+	alphaBeta := createIssue(t, env, pid, "alpha beta")
+	runCLI(t, env, dir, "label", "add", alphaOnly, "alpha")
+	runCLI(t, env, dir, "label", "add", alphaBeta, "alpha")
+	runCLI(t, env, dir, "label", "add", alphaBeta, "beta")
+
+	out := runCLI(t, env, dir, "list", "--label", "alpha", "--label", "beta")
+	assert.Contains(t, out, "alpha beta")
+	assert.NotContains(t, out, "alpha only")
+}
+
+func TestList_RepeatedNoLabelFiltersExcludeEveryLabel(t *testing.T) {
+	env, dir, pid := setupCLIWorkspace(t)
+	alpha := createIssue(t, env, pid, "alpha candidate")
+	beta := createIssue(t, env, pid, "beta candidate")
+	createIssue(t, env, pid, "plain candidate")
+	runCLI(t, env, dir, "label", "add", alpha, "alpha")
+	runCLI(t, env, dir, "label", "add", beta, "beta")
+
+	out := runCLI(t, env, dir, "list", "--no-label", "alpha", "--no-label", "beta")
+	assert.Contains(t, out, "plain candidate")
+	assert.NotContains(t, out, "alpha candidate")
+	assert.NotContains(t, out, "beta candidate")
+}
+
 // TestList_HumanRowUsesGlyphLayout pins the new row renderer's layout: an
 // open issue renders the open glyph "○ " ahead of its title, replacing the
 // old "%-8s  %-8s" column format.
--- a/cmd/kata/metadata_cli_test.go
+++ b/cmd/kata/metadata_cli_test.go
@@ -62,6 +62,18 @@ func TestList_MetaFilterPresence(t *testing.T) {
 	assert.NotContains(t, out, "unbranched")
 }
 
+func TestList_RepeatedMetaFiltersRequireEveryMatch(t *testing.T) {
+	env, dir, _ := setupCLIWorkspace(t)
+	_ = runCLI(t, env, dir, "--quiet", "create", "matching metadata",
+		"--meta", "team=core", "--meta", "tier=one")
+	_ = runCLI(t, env, dir, "--quiet", "create", "partial metadata",
+		"--meta", "team=core")
+
+	out := runCLI(t, env, dir, "list", "--meta", "team=core", "--meta", "tier=one")
+	assert.Contains(t, out, "matching metadata")
+	assert.NotContains(t, out, "partial metadata")
+}
+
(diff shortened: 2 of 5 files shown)


mdickinson:
Thanks for the quick fix!

PR #202, opened 2026-07-23
Automate federation enrollment from daemon config

wesm:
Federating a spoke currently requires an operator to coordinate enrollment and join steps or use the TUI for every project. This adds declarative `[[federation.project]]` mappings to `config.toml`, with a catalog hub, spoke and hub project names, and actor.

```toml
[[daemon]]
name = "team-hub"
url = "https://hub.example"
token_env = "KATA_TEAM_HUB_TOKEN"

[[federation.project]]
hub = "team-hub"
spoke_project = "spoke-project"
hub_project = "hub-project"
actor = "user-a"
```

With those sections in the spoke's config, restarting its daemon is enough to create or adopt `spoke-project`, enroll it against `hub-project`, and keep the mapping idempotent. There is no TUI or manual enroll/join handoff, and the local and hub project names can differ.

On daemon startup, a fail-open background reconciler creates or adopts the local project, ensures the hub project and project-scoped enrollment idempotently, honors token-identity actor precedence, and reports sanitized aggregate health without delaying listener readiness. Removing a mapping does not detach the replica; teardown remains an explicit `kata federation leave`.

Generated credentials are durably reserved under the stable hub project UID, while compatible manual credentials make a single local-to-hub UID transition during adoption. Hub authentication and redirects stay pinned to the configured origin, reverse-proxy path prefixes remain intact, and exact-token rotation converges every grant in scope.

Explicit leave uses a prepare/finalize boundary: the spoke durably marks a managed credential as leaving, drains earlier enrollment or rotation requests, records their enrollment IDs, and suppresses the mapping after teardown. A restart with an uncertain marker replays the reserved token to recover and revoke the exact enrollment. This prevents config reconciliation from stranding a hub grant between hub enumeration and local cleanup while preserving `--local-only` as the deliberate no-revoke escape hatch.

The API, generated client, changelog, configuration reference, and federation operations guide describe the resulting 0.7 contract; persisted database schemas are unchanged.


wesm:
working on this

PR #204, opened 2026-07-25
Route trusted Linux builds to managed runners

wesm:
## Summary

- route canonical same-repository Linux pull requests and main-branch builds to managed Linux runners, with forks and noncanonical copies falling back to GitHub-hosted runners
- invoke the reusable workflow from `main` and let checkout use the caller event's immutable SHA without accepting a dispatcher-supplied ref
- lint proposed workflow revisions on GitHub-hosted infrastructure

The managed runner group admits only the listed reusable workflow file from `main`; PR-controlled dispatcher changes therefore cannot target it directly. The reusable-workflow bootstrap is incorporated in this branch.

### DIFF
--- a/.github/workflows/test-pr.yml
+++ b/.github/workflows/test-pr.yml
@@ -1,18 +1,15 @@
 name: Tests
 
-# Fork PRs use GitHub-hosted runners. Same-repository PRs call the
-# main-branch workflow, whose Linux jobs can use the managed public fleet.
+# The PR dispatcher always calls the main-pinned workflow. That workflow
+# independently routes same-repository Linux jobs to the managed public fleet
+# and fork jobs to GitHub-hosted runners.
 on:
   pull_request:
-  pull_request_target:
 
 permissions: read-all
 
 jobs:
   run:
-    if: >-
-      (github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name != github.repository) ||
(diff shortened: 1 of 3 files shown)


wesm:
This is a false positive under the organization policy currently applied to these runners. I verified the live runner-group configuration through the GitHub API:

- workflow restriction is enabled (`restricted_to_workflows: true`)
- this repository is explicitly selected
- the only workflow in this repository permitted to use the group is `kenn-io/kata/.github/workflows/test.yml@refs/heads/main`

Therefore a fork can modify `test-pr.yml` and request `kenn-linux-x64-public`, but matching the label is not sufficient for admission. The PR-controlled workflow is not in the selected-workflow allowlist, so GitHub cannot dispatch that direct job to the managed group. Only the main-branch reusable workflow can use it, and that workflow independently checks the canonical repository plus both head and base repository identities before choosing the managed label; fork calls fall back to `ubuntu-latest`.

GitHub documents selected-workflow runner groups as the mechanism for restricting a group to a specific reusable workflow: https://docs.github.com/en/enterprise-cloud@latest/actions/how-tos/manage-runners/self-hosted-runners/manage-access

The finding would become valid if the selected-workflow restriction were removed or broadened, or if another runner group exposed the same label without an equivalent restriction. Neither is true in the live configuration.

PR #209, opened 2026-07-29
Open issue refs directly in the TUI

### LINKED ISSUE #208: Allow kata tui to open a known issue directly
`kata tui` always opens the issue browser. When a caller already has an issue ref from `kata list`, `kata create`, or another workflow, reaching its detail view still requires an extra search and selection step.

Accept an optional ref such as `kata tui abc4` to open the existing detail view directly, while preserving the current no-argument browser. The ref should accept the same short, qualified, and full UID forms as `kata show`.

salmonumbrella:
## Why

`kata tui` always opens the issue browser. Workflows that already have an issue ref from `kata list`, `kata create`, or another command still need an extra search and selection step to reach its details.

## Behavior

- `kata tui <issue-ref>` resolves the exact ref before Bubble Tea starts and opens the existing detail view directly.
- Bare short IDs, qualified short IDs, and full UIDs match the forms accepted by `kata show`.
- Lookup errors return before the alternate screen opens, and the issue list still loads behind the detail view for normal back navigation.
- CLI help, the README, quickstart, and CLI reference document the optional ref.

Closes #208

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@ kata create "fix login race"           # prints a short id, e.g. abc4
 kata list                              # see open work
 kata show abc4                         # inspect by short id
 kata tui                               # browse and triage interactively
+kata tui abc4                          # open an issue directly in the TUI
 ```
 
 `kata create` prints each issue's short id; use it in later commands. Close only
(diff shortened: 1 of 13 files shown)


wesm:
Nice UX improvement

PR #220, opened 2026-08-03
ready/next --all: compose with the scoped filters

### LINKED ISSUE #219: ready --all: support --label / --unowned / --owner for cross-project queue views
`kata ready --all` lists ready issues across every project, but rejects the filter flags outright: `--all does not support --unowned, --owner, --label, or --no-label`. I would like them to compose:

    kata ready --all --label handoff-to:macazbd --unowned

The output side already works — `ready --all` prefixes refs with the project (`issue=jibot-code#w6n5`), and cross-project refs resolve in `show` without `--project`. The filter flags are the only missing piece for a cross-project queue view.

## The use case: baton + buzz — an agent task-handoff lane on kata

Context, since it may interest you. We run kata as the shared ledger for a multi-machine agent fleet: the daemon lives on a small VM, and Claude Code / Codex / other harness sessions on several Macs are the clients. On top of it we built **baton**, a small CLI that passes tasks between machines and harnesses. A handoff is a kata issue carrying labels like `to:<host>`, `from:<host>`, `harness:<claude-code|codex|...>`, with a cold-start brief in the body written from the sender's session context. An agent scans the queue with `kata ready --label to:<host> --unowned`, claims with `kata claim`, and closes with typed evidence. The claim atomicity is doing real work for us — yesterday two concurrent sessions raced the same claim and it resolved cleanly, one winner, no double work.

The visible wire is [buzz](https://github.com/block/buzz): every handoff gets a thread in a #handoffs channel, and reactions mirror the ledger state (🎯 passed, 🙋 claimed, ✅ done, ↩️ bounced, ❌ dropped) so a human can watch the queue like a chat. kata stays authoritative — a buzz failure only warns, a kata failure aborts.

Today every handoff lives in a dedicated `handoffs` project, purely because per-label queue views are project-scoped. That forces a two-row model: an envelope in `handoffs` plus the real work issue in its home project, joined by a `related` link — and the two drift. We caught an envelope sitting claimed-but-open after its home issue was already closed. We want to collapse to one row — handoff labels directly on the work item in its home project — and `ready --all --label` is the missing queue view. With ~38 projects on our daemon, emulating it client-side is 38 round-trips per scan.

Possibly related to the direction in #4 (saved views / smart filters); plain flag composition on `ready --all` would already cover us.

Happy to send a PR if you are open to this. Is there anything in the current `--all` implementation that makes label filtering awkward, or flag semantics you would prefer?

Joi:
Closes #219.

`kata ready --all` listed ready issues across every project but rejected `--unowned`, `--owner`, `--label`, and `--no-label`, so a cross-project filtered queue view — e.g. every unowned ready issue labeled `handoff-to:example-host`, anywhere — required one `ready` call per project. With ~40 projects on a daemon that is 40 round-trips for one queue scan.

## What changed

- `ReadyGlobalRequest` gains `unowned`, `owner`, `label`, and `exclude_label` query parameters; the global handler validates `unowned`/`owner` mutual exclusion exactly like the per-project endpoint and passes a `ReadyIssuesFilter` through.
- `ReadyIssuesGlobal` in both stores extends the cross-project query with the same filter clauses as `ReadyIssues` (AND logic, case-insensitive labels), keeping ordering and readiness semantics identical.
- The client-side guard is removed; `next --all` inherits the filters through the shared options. `--all` remains mutually exclusive with `--project`.
- Store conformance exercises the filters against both backends across two projects; OpenAPI artifacts and the generated client are refreshed via `make api-generate`; `docs/reference/cli.md` reflects the composed behavior.

Context, for the curious: this serves an agent task-handoff lane where kata is the shared ledger for a multi-machine fleet and a chat channel mirrors claim state — described in #219. With this change the dedicated single-project queue workaround can collapse into labels on the work items themselves.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -6307,6 +6307,34 @@ paths:
           schema:
             format: int64
             type: integer
+        - explode: false
+          in: query
+          name: unowned
+          schema:
+            type: boolean
+        - explode: false
+          in: query
+          name: owner
+          schema:
+            type: string
+        - explode: true
+          in: query
+          name: label
+          schema:
+            items:
+              type: string
+            type:
+              - array
+              - "null"
+        - explode: true
+          in: query
+          name: exclude_label
+          schema:
+            items:
+              type: string
+            type:
+              - array
+              - "null"
       responses:
         "200":
           content:
--- a/cmd/kata/next_test.go
+++ b/cmd/kata/next_test.go
@@ -237,10 +237,6 @@ func TestNext_AllValidationMirrorsReady(t *testing.T) {
 		want string
 	}{
 		{name: "project", args: []string{"--project", "anything", "next", "--all"}, want: "mutually exclusive"},
-		{name: "unowned", args: []string{"next", "--all", "--unowned"}, want: "--all does not support"},
-		{name: "owner", args: []string{"next", "--all", "--owner", "alice"}, want: "--all does not support"},
-		{name: "label", args: []string{"next", "--all", "--label", "bug"}, want: "--all does not support"},
-		{name: "no-label", args: []string{"next", "--all", "--no-label", "wip"}, want: "--all does not support"},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
@@ -250,6 +246,17 @@ func TestNext_AllValidationMirrorsReady(t *testing.T) {
 			assert.Contains(t, err.Error(), tt.want)
 		})
 	}
+}
+
+func TestNext_AllAcceptsFilterFlags(t *testing.T) {
+	// next shares ready's option validation: the scoped filters compose with
+	// --all so an agent can pick, e.g., the next unowned issue anywhere.
+	env, dir, pid := setupCLIWorkspace(t)
+	createIssue(t, env, pid, "alpha")
+
+	out, err := runCLICapture(t, env, dir, "next", "--all", "--unowned")
+	require.NoError(t, err)
+	assert.Contains(t, out, "alpha")
 }
 
 func TestNext_UnownedAndOwnerAreMutuallyExclusive(t *testing.T) {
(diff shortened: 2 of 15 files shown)


wesm:
thank you!

Joi:
Field note from the use case that motivated #219, after building against it: **`ready --all` composition alone doesn't cover the queue-scan case, because `ready` excludes blocked issues.**

The workload is "every issue labeled `handoff-to:<host>`, anywhere" — a cross-project queue view. I assumed `ready --all --label` was the primitive, built the per-project fan-out as a stopgap, and then found the fan-out had to use `list`, not `ready`: `ready` means *open and unblocked*, so any handoff pointing at a blocked issue silently vanishes from the queue. Measured on our daemon, one project (`jibot-code`) has 124 open issues and 115 ready — 9 would have been invisible, with no warning. For a work queue that's a correctness bug, not a filtering preference: a blocked item is still real work someone is waiting on, it just isn't startable yet.

`list` has no `--all` at all today, so the fan-out stays regardless of whether this PR merges: 37 projects × ~354ms = ~14s sequential, ~1.7s at 8-way, against ~0.7s for a single scoped call. That lands on an interactive path for us (a session-start hook), which is what makes it worth raising rather than just absorbing.

Suggestion: give `list` the same `--all` + `unowned`/`owner`/`label`/`exclude_label` composition this PR adds to `ready`/`next`, either here or as a follow-up. It looks like the same shape — `ListGlobalRequest` mirroring `ReadyGlobalRequest`, same mutual-exclusion validation, same conformance coverage — and it would make "cross-project filtered queue" work for queues whose items can be blocked.

Happy to extend this PR with that if you'd rather have it in one change than two. No urgency on our side — we have a local index planned for the hot path either way.

Joi:
Apologies — my note above landed about an hour after you'd already merged, so "happy to extend this PR" was stale by the time I wrote it. Filed properly as #221 instead.

Issue #222, opened 2026-08-04
ready --all: new client + old daemon silently ignores label/unowned filters (returns everything)


Joi:
Following up on #220 (thanks for merging). Testing it before a fleet upgrade turned up a silent-wrong-answer case in client/daemon version skew that I think is worth guarding.

## What happens

A client built from `main` (with #220) talking to a **v0.13.0 daemon** sends the new `label` / `unowned` / `owner` / `exclude_label` query parameters on the global endpoint. The old daemon doesn't know them, ignores them, and returns **everything** — with a 200 and no warning.

Measured against our daemon:

```
# new client -> old (v0.13.0) daemon
$ kata ready --all --label handoff-to:hosta --agent
OK ready count=427          # every ready issue in the fleet

# same client -> daemon built from the same commit
$ kata ready --all --label handoff-to:hosta --agent
OK ready count=2            # correct
```

## Why it matters more than a normal skew bug

The failure direction is "too many results," silently. For the label-scoped queue views #219/#220 were filed to enable, that's the worst possible shape: a work queue that should show 2 items shows 427, all of them unrelated issues, and nothing indicates the filter was dropped. A consumer that acts on the first row acts on the wrong issue.

It also can't be caught by the operator, because both the exit code and the output format are exactly what success looks like.

## Suggestion

The responses already carry `kata_api_version`. Two options, either would do:

- Have the global handler **reject** requests carrying filter params it doesn't understand (fail closed), so an old daemon returns a clear error instead of unfiltered rows.
- Or have the client compare `kata_api_version` against what the requested flags need and refuse locally with something like "`--all --label` requires daemon API vN; this daemon reports v1 — upgrade the daemon."

The second is nicer for rollout because it names the fix.

No urgency for us — we upgrade the daemon and clients together as a gated step, and this is now written into our runbook. Filing it because the silent direction seems worth closing for anyone who rolls clients out first, which is the more natural order.

PR #227, opened 2026-08-06
search: support --label / --no-label for scoped lookups

### LINKED ISSUE #225: search: support --label / --no-label for scoped lookups
Adjacent to the #219/#220/#221 label-filter thread, but on `search` rather than `list`/`ready`.

## What happens

`search` takes no attribute filters:

```
$ kata search "someday ready filter" --project example-project --label friction
ERR search usage: unknown flag: --label
```

The complement is missing too: `list` filters by label but takes no query text. So "find issues about X that are labeled Y" isn't expressible in one command.

## Why it matters

This is the dedup shape. Agent guidance says to search before creating, and when the target is a labelled subset of a project, neither verb answers it:

- `search "<terms>"` ranks across the whole project — for a 4-word query, hybrid mode returned 4 unrelated issues at score 0.02, none of them in the labelled subset.
- `list --label <l>` returns the subset but can't rank it, so the agent reads the whole label.

Fine at 2 items. Not fine at 200, which is where the dedup check quietly stops working and duplicates start landing.

## Ask

`--label` (repeatable, AND) and `--no-label` on `search`, matching `list` semantics and #189's multi-instance fix.

## One wrinkle

The lexical leg can filter in SQL. The vector leg retrieves top-K by similarity first, so filtering after retrieval would silently under-return — `--limit 20 --label x` yielding 2 rows while 50 match. Same silent direction as #222. Over-fetching with a documented ceiling plus the existing `degraded`/`degraded_reason` signal would cover it; pre-resolving the candidate set would be exact but heavier.

Happy to implement whichever you prefer.

salmonumbrella:
Closes #225.

`kata search` took no attribute filters, so the dedup shape — "find issues about X that are labeled Y" — wasn't expressible in one command: `search` ranked across the whole project while `list --label` returned the subset but couldn't rank it. Fine at 2 items, not at 200.

## What changed

- `SearchFTS`/`SearchFTSAny` take a `SearchFTSParams` struct carrying `Labels`/`ExcludeLabels`; both stores push the same `EXISTS`/`NOT EXISTS` predicates `list` uses into the FTS query before `LIMIT`, so the lexical leg filters exactly and never under-returns (AND logic, case-insensitive labels). Store conformance exercises both backends, including a case that pins filter-before-LIMIT on both tie-break orders.
- The vector leg filters during hit hydration and reuses the existing one-shot deep retry when a full batch filters down short. When filters are present and the deep batch comes back full with fewer than `limit` survivors, the response carries the existing `degraded`/`degraded_reason` signal with a fixed reason — on the success path, so an explicit `--hybrid`/`--semantic` request is never turned into a 503 by the ceiling. Unfiltered searches behave identically to before.
- `SearchRequest` gains `label`/`exclude_label` query params with `,explode` tags, the explode regression guard from the #189 fix now covers them, and OpenAPI artifacts are refreshed via `make api-generate`. No `api_schema_version` bump, matching #220's precedent.
- `kata search` gains `--label`/`--no-label` (`StringSliceVar`, same help text and semantics as `list`), with endpoint, CLI round-trip, and e2e coverage for repeated-flag AND, exclusion, and case-insensitivity.

Same silent direction as #222 applies during version skew: an old daemon ignores the new params and returns unfiltered results. Nothing here changes that class either way — noting it since search now adds two more params to it.


wesm:
looking at this

PR #232, opened 2026-08-07
Expose standard daemon pprof endpoints

mariusvniekerk:
## Summary

Kata operators need a supported way to inspect CPU, memory, goroutine, mutex, blocking, and trace behavior in a running daemon with standard Go tooling.

This exposes the standard `net/http/pprof` handlers through the daemon's existing listener and documents profile capture for remote daemon operators. The profiling surface is intentionally available without Kata API or browser-session authentication; listener reachability is its access boundary.

<sup>generated by a clanker</sup>

### DIFF
--- a/docs/operations/remote-daemon.md
+++ b/docs/operations/remote-daemon.md
@@ -30,6 +30,31 @@ set once.
 Run the daemon under a process manager such as launchd, systemd, or a container
 runtime on the host that owns the SQLite database.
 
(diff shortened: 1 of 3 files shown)


mariusvniekerk:
kata is a local tool.  The pprof endpoints are accessible BY DESIGN.  Rejected your stupid hardening bullshit.

PR #237, opened 2026-08-08
list: add cross-project filters and guard version skew

### LINKED ISSUE #221: list: add --all with the same filter composition as ready --all (#220)
Follow-up to #219 / #220 (thanks for the quick merge on that one).

#220 gave `ready --all` / `next --all` composition with `--unowned`/`--owner`/`--label`/`--exclude-label`. Building against it surfaced that the cross-project **queue scan** case needs the same thing on `list`, and `list` has no `--all` today.

## Why `ready --all` isn't sufficient

`ready` means *open and unblocked*. That's the right semantics for "what can I start now," but a work queue also has to show items that are real and waiting but not yet startable — otherwise they vanish silently rather than appearing as blocked.

Concretely: we label issues `handoff-to:<host>` to hand work between machines, and scan for that label across all projects. On our daemon one project has 124 open issues and 115 ready. Using `ready` for the scan means 9 handoffs are invisible — no row, no warning, no indication anything was filtered. For a queue view that reads as "no work waiting," which is worse than a slow answer.

So the scan has to use `list --label ... --status open`, which means one call per project: ~40 round-trips at ~354ms each ≈ 14s sequential, ~1.7s at 8-way. Exactly the cost #219 was filed to remove, just moved to a different verb.

## Ask

Give `list` an `--all` that composes with the same filters #220 added to `ready`/`next` — ideally `--label` (repeatable, AND), `--exclude-label`, `--unowned`/`--owner` (same mutual-exclusion validation), and `--status`.

Two details from using it that might be worth folding in:

- `list`'s `--limit` defaults to 200 and `--limit 0` returns **zero** rows, while on `ready` 0 means unlimited. That asymmetry is easy to trip over — a scan that looks complete but silently truncates at 200 is the same class of failure as the blocked-issue one above. Worth either aligning the zero semantics or documenting the difference where `--all` is introduced.
- If `--all` on `list` lands, a global result is more useful with the project on each row, since short ids are only unique per project.

Happy to send a PR for this if it's welcome — it looks like the same shape as #220 (a `ListGlobalRequest` mirroring `ReadyGlobalRequest`, same validation, same conformance coverage across both stores). Say the word and I'll pick it up; equally happy to leave it to you if you'd rather keep the global-query surface consistent by doing it yourself.

No urgency — we cache the hot path locally, so this is a simplification for us, not a blocker.

### LINKED ISSUE #222: ready --all: new client + old daemon silently ignores label/unowned filters (returns everything)

salmonumbrella:
Closes #221.
Closes #222.

`ready --all` and filtered search can send query parameters that older daemons silently ignore, while `list` has no cross-project queue scan. That makes fleet work queues either incomplete or silently unfiltered.

## What changed

- Add `kata list --all` over the existing global issue feed, with status, priority, owner/unowned, repeatable label, exclusion, and metadata filters composing in one request.
- Return qualified issue refs in human and agent output, and include `project_name` in global JSON rows.
- Make `--limit 0` mean unlimited and leave `list --all` unlimited by default, while preserving the per-project default of 200.
- Push the same filter semantics through SQLite and PostgreSQL, including case-insensitive label matching and AND composition.
- Bump the API schema to `0.9.0` and regenerate the Go and browser clients for the new global response shape.
- Preflight filtered global ready/next, filtered search, and filtered global list requests against the daemon API version so old daemons fail closed with an upgrade message.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -2244,6 +2244,18 @@ components:
         set_parent:
           type: string
       type: object
+    ListAllIssuesResponseBody:
+      additionalProperties: true
+      properties:
+        issues:
+          items:
+            $ref: "#/components/schemas/ListGlobalIssueOut"
+          type:
+            - array
+            - "null"
+      required:
+        - issues
+      type: object
     ListFederationEnrollmentsBody:
       additionalProperties: true
       properties:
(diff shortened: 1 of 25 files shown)


wesm:
rebasing

PR #258, opened 2026-08-11
Recommend Homebrew Core for macOS installs

### LINKED ISSUE #213: Add Homebrew installation
Kata already publishes versioned macOS release artifacts, but it is not currently installable through Homebrew. Could Kata be added to Homebrew Core so macOS and Linux users can install and upgrade it with:

```sh
brew install kata
```

[msgvault is already available in Homebrew Core](https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/m/msgvault.rb), so it may be a useful precedent for the formula and release workflow.

wesm:
## Why

Kata is now available directly from Homebrew Core. macOS users no longer need the tap-qualified formula name, so the public installation path should use the shorter Core command.

## What reviewers should know

The README and website now recommend `brew install kata` on macOS. Linux keeps the release installer as its primary path, and Windows remains unchanged. Historical release notes and the packager-facing tap contract remain intact.

Closes #213.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,10 @@ Linux, or macOS without Homebrew:
 curl -fsSL https://katatracker.com/install.sh | bash
 ```
 
+Linux and WSL 2 users who already use
+[Homebrew](https://docs.brew.sh/Homebrew-on-Linux) can also run
+`brew install kata`.
+
 Windows PowerShell:
 
 ```powershell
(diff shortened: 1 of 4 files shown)


chrisfinazzo (commented):
You can also add that Homebrew is [available][] for Linux and via WSL, so kata should work there as well.

[available]: https://docs.brew.sh/Homebrew-on-Linux

PR #261, opened 2026-08-11
Expand native MCP server coverage

### LINKED ISSUE #257: mcp: add project mode (parity with list/ready/next --all)
## The situation

`kata mcp serve` binds one project and one actor when the process starts, and
none of its 13 tools accepts a project input. I wrote that server in #238, and
the binding was a deliberate call — #3 choice 4(a), taken over per-call
targeting because an agent could write to the wrong project.

I think the call was wrong, for a reason that only became visible once the
server was in use: it leaves MCP as the only Kata surface that cannot work
across projects.

Everything else already can. `list --all`, `ready --all`, and `next --all` span
non-archived projects. The daemon exposes `listAllIssues`, `readyIssuesGlobal`,
and `resolveProject`. `AGENTS.md` documents `project#ref` as the cross-project
reference form, and the daemon resolves it — a qualified ref even wins over a
conflicting `--project`:

```console
$ kata --project hub-project show spoke-project#abc4
# returns the spoke-project issue
```

MCP is the one surface where that vocabulary does not exist.

## What works today, and why it is not enough

Running one server process per project works, and for a repository-scoped
coding agent it is the correct default. That agent is already scoped to one
repository, so a project-bound server matches its context exactly. Nothing here
should change that default.

It stops working for an assistant-style client: one connection, one agent, a
whole daemon's worth of work to triage. That client cannot answer "what is ready
anywhere" without the host launching a process per project and merging the
results itself — which is the work `ready --all` already does inside the daemon.
On a daemon with a dozen projects that is a dozen long-lived processes and a
dozen host entries, to get what the CLI gets from one flag.

## Ask

An opt-in cross-project mode. The current behavior stays the default.

```sh
kata mcp serve --workspace PATH               # unchanged default
kata mcp serve --all-projects                 # every non-archived project
kata mcp serve --projects spoke-project,hub-project
```

In that mode:

- Reads (`kata.list`, `kata.ready`) take an optional `project`. Omitted means
  every project in scope, matching `--all`.
- `kata.show` accepts `project#ref` in its existing `ref` field.
- Writes require an explicit target — `project` for `kata.create`,
  `project#ref` for every tool that takes a ref. There is no ambient project to
  fall back to.
- A new read-only `kata.projects` lists projects in scope. Without it an agent
  cannot discover a legal target.
- The actor still binds at startup. #3 choice 5 is untouched; tool inputs still
  take no actor and one agent still cannot forge another's identity.

## Why per-call targeting is safer than it looked in #3

The objection in choice 4 was that an agent could write to the wrong project.
That is a real hazard, but it belongs to the option as it was framed there — a
`workspace` field on every tool, with an implicit default when omitted.

Requiring qualification on writes removes the default, and with it the hazard.
Compare the two failure modes:

- **Bound mode**: the target is ambient. It comes from a `--workspace` path or a
  host config file the agent never sees. If that binding is wrong, every write
  lands in the wrong project, the tool call looks correct, and nothing in the
  transcript shows the mistake.
- **Explicit mode**: the target is named at the call site. A wrong target is
  visible in the tool call, in the transcript, and in the approval prompt of any
  host that gates tool calls.

`kata.close` on `spoke-project#abc4` states what it is closing. `kata.close` on
`abc4` does not.

Reads are not exposed to this and should default to all projects in scope, the
same as `--all`.

## Disambiguation is already solved

`docs/reference/mcp.md` gives the reason relationship inputs reject bare
identifiers: an unscoped full UID does not prove project membership, and a
26-character short ID cannot be distinguished from a globally resolved ULID.

`project#ref` avoids both. Neither a ULID nor a bare short ID contains `#`, so
the qualified form is unambiguous by construction. It needs no new parsing
rules — it is what `list --all` prints, what `AGENTS.md` documents, and what the
daemon already resolves. MCP tool results already carry `qualified_ref` next to
`ref`, so the output half of this shipped in #238.

## One genuine gap

`kata search` has no `--all`, and the daemon's search is project-scoped. Cross-
project search therefore needs either a global search endpoint or client-side
fan-out against the existing eight-concurrent-call ceiling.

I would keep `kata.search` project-scoped until a global endpoint exists, so
this stays an MCP change and does not quietly become a daemon API change.

## Out of scope

- **HTTP transport.** #3 choice 1(a) kept the server stdio-only because HTTP
  reopens the LAN-exposure question the README defers. Separate decision, and
  cross-project mode is useful to stdio clients without it.
- Cross-project relationship links. Different semantics, different issue.
- The operations #238 deliberately excludes: delete, purge, restore, project
  administration, tokens, federation, sync, import, export, recurrence, event
  streaming, digest, force-create, force-claim.

## Unrelated small thing

`docs/reference/mcp.md` describes `kata.list` as "List compact issue summaries
with project filters." The runtime schema says "in the bound project" and has no
project parameter — that reads as though cross-project support already partly
exists. Happy to fold the correction into whatever lands here, or send it
separately.

salmonumbrella:
## What changed

- Makes bare `kata mcp serve` follow every project visible to the selected daemon. `--workspace`, `--project`, and `--projects` remain available when a client needs a narrower boundary; the redundant `--all-projects` setting is gone.
- Starts with 13 read-only section loaders instead of placing all 53 detailed tools in the initial model context. Loading a section registers its typed tools and publishes the standard tool-list change notification, while each operation keeps its own schema and safety annotations.
- Expands coverage across issue workflows, project and token administration, federation, GitHub sync, recurrence, import, digest, and events. Two JSONL storage tools remain conditional on operator startup configuration.
- Adds first-class `scheduled_on` and timezone inputs, `someday=true` parking with null removal, force-create, and force-claim support.
- Preserves daemon authorization, revision checks, destructive confirmations, origin-pinned credentials, bounded results, and secret redaction. This adds no database schema or migration change, and host process control remains outside MCP.

## Why

MCP was the only Kata client surface fixed to one project, and it exposed only a small part of the daemon. Assistant clients had to start one server per project and carry dozens of unrelated tool definitions before they could do normal administration or workflow work.

The daemon already defines the accessible project catalog. The MCP process now follows that catalog by default, while progressive section loading keeps the initial tool surface small without replacing typed operations with a generic command tool.

## Usage

```sh
# Every project visible to the selected daemon
kata mcp serve

# One project
kata --workspace /path/to/repository mcp serve

# Fixed project allowlist
kata mcp serve --projects example-project,shared-project
```

Call the applicable `kata.load_*` section tool before using its detailed tools. Multi-project issue references use `example-project#abc4`. Scheduling accepts a date, a local date-time with an IANA timezone, or a UTC RFC 3339 value ending in `Z`. Native parking uses `{"metadata":{"someday":true}}`; set the key to JSON `null` to remove it.

Host-storage import and export stay disabled unless the operator opts in:

```sh
kata mcp serve \
  --storage-root /srv/kata/exchange \
  --storage-target restore=restore.db
```

Closes #257

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,9 +13,10 @@ Production Postgres deployments can use separate schema-owner and runtime roles;
 see the [operator ceremony](docs/operations/postgres.md).
 Go applications can instead mount kata's listener-free HTTP service in-process;
 see [Embedding kata in Go](docs/development/embedding.md).
-MCP clients can start Kata's native stdio server with `kata mcp serve`. It can
-use the complete daemon catalog by default, or narrow access to one project or
-a fixed allowlist. Thirteen section loaders progressively expose Kata's typed
+MCP clients can start Kata's native stdio server with `kata mcp serve`. It
+binds the current workspace's project by default, and can serve a fixed
+allowlist or the complete daemon catalog on request. Thirteen section loaders
+progressively expose Kata's typed
 issue, administration, automation, and event workflows. See the [MCP
 reference](docs/reference/mcp.md).
 
--- a/README.md
+++ b/README.md
@@ -13,8 +13,11 @@ Production Postgres deployments can use separate schema-owner and runtime roles;
 see the [operator ceremony](docs/operations/postgres.md).
 Go applications can instead mount kata's listener-free HTTP service in-process;
 see [Embedding kata in Go](docs/development/embedding.md).
-MCP clients can start Kata's native, project-bound stdio server with `kata mcp
-serve`; see the [MCP reference](docs/reference/mcp.md).
+MCP clients can start Kata's native stdio server with `kata mcp serve`. It can
+use the complete daemon catalog by default, or narrow access to one project or
+a fixed allowlist. Thirteen section loaders progressively expose Kata's typed
+issue, administration, automation, and event workflows. See the [MCP
+reference](docs/reference/mcp.md).
 
 The documentation in [`docs/`](docs/) is the definitive guide, published with
 Zensical at <https://katatracker.com/>.
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -85,6 +85,9 @@ components:
       properties:
         actor:
           type: string
+        event_id:
+          format: int64
+          type: integer
         evidence_types:
           items:
             type: string
@@ -112,6 +115,7 @@ components:
         - actor
         - issue
         - reason
+        - event_id
       type: object
     AuditClosesResponseBody:
       additionalProperties: true
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -4237,7 +4237,7 @@ components:
       type: object
 info:
   title: kata
-  version: 0.10.0
+  version: 0.11.0
 openapi: 3.1.0
 paths:
   /api/v1/audit/closes:
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -417,6 +417,8 @@ components:
       properties:
         incoming:
           type: boolean
+        to_project_uid:
+          type: string
         to_ref:
           type: string
         type:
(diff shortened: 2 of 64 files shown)


wesm:
looking

wesm:
working still

PR #262, opened 2026-08-11
Make fast SQLite mode eliminate durability flushes

cpcloud:
Windows tests still performed SQLite durability flushes with `KATA_TEST_FAST_SQLITE` enabled. The main database entered WAL before `synchronous=OFF`, while the vector sidecar ignored the setting.

This applies the guarded test settings before WAL in both stores and adds a vector regression test. No timeout increase is included.

Production binaries continue to ignore the environment variable and retain their existing durability behavior.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -181,7 +181,7 @@ jobs:
         shell: bash
         env:
           KATA_TEST_FAST_SQLITE: "1"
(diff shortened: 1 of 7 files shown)


cpcloud:
Merging. This doesn't address the flake, which seems very much like a Windows hiccup, but it should help mitigate the issue when hiccups cause very slow disk writes.

PR #266, opened 2026-08-12
Report unreachable local daemon errors

codyw912:
## Summary

- preserve local daemon probe errors when a live runtime PID exists
- avoid starting a redundant daemon when the recorded endpoint is inaccessible
- report the PID, endpoint, and underlying connection failure to the caller

## Motivation

A sandboxed client can sometimes read the daemon runtime metadata while being
unable to access its Unix socket. Previously, health checks reported that no
daemon was running and normal commands could attempt another start, eventually
reporting a startup timeout.

This distinguishes an absent daemon from a live but unreachable one while
preserving the existing `daemon_unavailable` CLI error category.


wesm:
looking at this, I think I just fixed this bug in kit (Go library we share between a bunch of projects) but I might be wrong

wesm:
will merge pending CI

PR #270, opened 2026-08-14
Add conventional kata --version flag

### LINKED ISSUE #269: kata --version is not accepted
## The situation

`kata --version` does not work:

```console
$ kata --version
kata: unknown flag: --version
$ echo $?
2
```

`--version` is the reflex probe for a CLI. It is what install checks, CI
pins, bug reports ("what version are you on"), and agents shelling out to
identify a binary all reach for first. Today every one of those gets a
usage error and exit 2 instead of a version.

## What works today, and why it is not enough

`kata version` works and already carries everything needed, including
machine-readable output:

```console
$ kata version --json
{"kata_api_version":1,"agent_format":1,"name":"kata","version":"v0.14.1", ...}
```

`docs/get-started/install.md` uses it as the post-install check, so the
documented path is fine once you know it. The gap is the guess: people and
scripts try `--version` first, and the failure reads like a broken binary
rather than a different spelling. Cobra hands `--version` to most tools for
free, which makes its absence here more surprising, not less.

## Ask

Accept `kata --version` as an equivalent spelling of `kata version`, with
the same output and the same `--json` / `--agent` handling.

Deliberately out of scope: a `-v` shorthand. `-v` conventionally means
verbose, and kata already exposes `-q` / `--quiet`, so leaving `-v` unbound
keeps it available for a future `--verbose`.

salmonumbrella:
## What changed

- `kata --version` prints the same build identity as `kata version` and honors `--json` and `--agent`.
- No `-v` shorthand, so that spelling stays free for a future `--verbose` alongside the existing `-q` / `--quiet`.

## Why

`kata --version` is the first thing a person or a script tries when it needs to know which binary is installed, and it failed with `unknown flag: --version` and exit 2. That reads like a broken binary rather than a different spelling, and it breaks install checks, CI pins, and agents that identify a tool before using it. `kata version` already carried the full answer; this makes the conventional spelling reach it.

## Usage

Real output from a build of this branch:

```console
$ kata --version
kata v0.14.1-1-g983f7ca
  commit:  983f7ca
  built:   2026-08-14T13:30:20-07:00
  go:      go1.26.6
  os/arch: darwin/arm64

$ kata --version --json
{"kata_api_version":1,"agent_format":1,"arch":"arm64","built":"2026-08-14T13:30:20-07:00","commit":"983f7ca","go":"go1.26.6","name":"kata","os":"darwin","version":"v0.14.1-1-g983f7ca"}

$ kata --version --agent
OK version version=v0.14.1-1-g983f7ca agent_format=1
```

`kata version` is unchanged.

Cobra's built-in `Version` field is deliberately not used: its generated flag short-circuits before `PersistentPreRunE` resolves the output mode, which would leave `kata --version --json` printing human text.

Closes #269

### DIFF
--- a/cmd/kata/main.go
+++ b/cmd/kata/main.go
@@ -27,6 +27,7 @@ type globalFlags struct {
 	Workspace    string
 	Project      string
 	Daemon       string
+	Version      bool
 }
 
 var flags globalFlags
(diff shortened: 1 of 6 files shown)


wesm:
rebasing

PR #271, opened 2026-08-15
Add a contract output mode for agent guidance

### LINKED ISSUE #267: Provide a non-repository mechanism for loading Kata agent guidance from `kata init --with-agents`
## Summary

Would you consider exposing the managed agent briefing installed by `kata init --with-agents` as a stable, marker-free CLI output mode?

```sh
kata quickstart --format contract
```

The output should be the same canonical briefing body currently written between:

```html
<!-- BEGIN KATA (managed by `kata init --with-agents`) -->
...
<!-- END KATA -->
```

but without those file-management markers.

## Motivation

`kata init --with-agents` is a good opt-in for repositories that want to adopt Kata as a shared, committed agent-workflow convention.

I am interested in using Kata locally in a shared repository, without:

* Modifying or committing `AGENTS.md`, `CLAUDE.md`, or other repository instruction files.
* Implying that all collaborators should install or use Kata.
* Creating a second, locally maintained copy of Kata’s operating policy that can drift from the installed Kata version.

Several agent harnesses have user-local ways to load instructions at session start:

* **Claude Code:** a user-level `SessionStart` hook can inject a command’s stdout or structured `additionalContext` into the current session.
* **Codex CLI:** its `SessionStart` hooks can likewise add context.
* **OpenCode:** supports user-local global instructions and configurable instruction sources.

A stable command that emits the managed Kata briefing would let users integrate the same policy locally in each of these environments.

For example, a Claude Code user-level hook could run:

```json
{
  "hooks": {
    "SessionStart": [
      {
        "matcher": "startup|resume|clear|compact",
        "hooks": [
          {
            "type": "command",
            "command": "kata quickstart --format contract"
          }
        ]
      }
    ]
  }
}
```

This would supply Kata’s policy to the agent at session start without changing anything in the repository.

In fact, a user-local integration can be installed globally but remain inert outside Kata workspaces. For example, a session-start hook can first detect the workspace binding:

```sh
test -f "$CLAUDE_PROJECT_DIR/.kata.toml" &&
    kata quickstart --format contract --workspace "$CLAUDE_PROJECT_DIR"
```

This injects Kata guidance only when the repository has explicitly opted into Kata via `.kata.toml`; unrelated repositories receive no additional context. It also means the harness integration itself does not need to be committed or configured per repository.

## Why this should be distinct from existing formats

`kata quickstart` / `kata agent-instructions` already provides the fuller agent operating contract. The `--with-agents` block is a different, shorter briefing:

* It directs agents to the full `kata quickstart` contract.
* It provides the most important CLI workflow rules.
* It includes `work.*` attention/orchestration conventions.
* It is specifically designed as persistent agent guidance.

That makes `contract` a semantic output mode, and I think it could fit in with the existing `--format` options

| Format | Intended output |
|---|---|
| `human` | Interactive terminal-oriented output |
| `agent` | Concise line-oriented command output for agent transcripts |
| `json` | Structured machine-readable command result |
| `contract` | Canonical managed agent briefing, suitable for injection into an agent session |

This is just a suggestion, though; you may prefer another subcommand or flag.

## Proposed behavior

```sh
kata quickstart --format contract
# alias
kata agent-instructions --format contract
```

This should:

* Write only the marker-free managed briefing body to stdout.
* Avoid terminal framing, progress messages, or workspace mutation.
* Work even when the current directory has not been initialized as a Kata workspace, since this is policy text rather than project state.
* Be documented as an automation-friendly API for agent harness integrations.
* Preserve the existing `--project` and `--workspace` selectors for contract output, so a harness can inject a contract tailored to an explicitly selected Kata project or workspace:

  ```sh
  kata quickstart --format contract --project my-project
  kata quickstart --format contract --workspace /path/to/project
  ```

  This leaves room for the contract renderer to include project-specific guidance where appropriate, while retaining a stable local-integration API.

Most importantly, `kata init --with-agents` and `kata quickstart --format contract` should use the same underlying template/renderer, so the static file-managed briefing and dynamically injected briefing cannot drift.

salmonumbrella:
## What changed

- Injects a graphviz DOT diagram into context.
- `kata quickstart --format contract` prints the managed agent briefing marker-free, with no terminal framing and no workspace mutation. It works in a directory that was never `kata init`ed, since this is policy text rather than project state.
- The `agent-instructions` alias and the existing `--project` and `--workspace` selectors all work with it, so a harness can point at an explicitly selected workspace.
- `contract` is scoped to `quickstart`. Every other command still rejects it with exit 2, and it conflicts with `--json` and `--agent` the same way the other modes do.
- The block body that `kata init --with-agents` writes between its markers moved into a single constant that both surfaces render, so the committed briefing and the injected one cannot drift.
- A hidden `agent-contract-hook` wraps that same text in Codex's structured SessionStart response, and `kata init --with-codex-hooks` now installs it next to the existing `work.attention` hook.

## Why

Adopting Kata in a shared repo currently means committing to `AGENTS.md` or `CLAUDE.md`, which signals that every collaborator should install Kata. The alternative is keeping a hand-maintained copy of Kata's operating policy in a user-level hook, and that copy drifts from the installed version the moment either one changes.

Claude Code, Codex, and OpenCode all have user-local ways to load instructions at session start. They just need a stable command that prints the briefing. This gives them one, and it stays correct because it is the same constant `--with-agents` writes.

## What gets injected

```mermaid
flowchart TB
  arrive{"Work arrives"} --> search["Search first; reuse an open issue<br/>or create one"]
  search --> route{"Work it, or delegate it?"}

  subgraph workside["Working a kata-tracked issue"]
    direction TB
    claim["On claim or start, mark it actively tracked:<br/>kata meta set #60;ref#62; work.attention ok"]
    branch["If the work happens on a dedicated branch, stamp it once:<br/>kata meta set #60;ref#62; work.branch #60;branch#62;<br/>or bind at creation:<br/>kata create ... --meta work.branch=#60;branch#62; --idempotency-key #60;key#62;"]
    live["Keep your live state truthful on the issue:<br/>kata meta set #60;ref#62; work.attention stuck#124;needs-human#124;ok<br/>with a one-line kata meta set #60;ref#62; work.attention_msg #34;#60;why#62;#34;<br/>Raise stuck when you cannot proceed, needs-human when you want<br/>input or review (you may keep working), and clear back to ok when unblocked"]
    claim --> branch --> live
  end

  subgraph delegateside["Delegating work as separate issues (fan-out/join)"]
    direction TB
    fanout["Create each sub-issue with --meta work.branch=...<br/>and an idempotency key; capture refs from --json (.issue.short_id)"]
    join["Join with kata wait #60;refs#62; --until attention --any<br/>Matches needs-human or stuck; a close also completes the wait,<br/>and the reported reason distinguishes which. Use --timeout so a<br/>wrapper can tell timeout from satisfaction."]
    coord["As coordinator you read work.* —<br/>you never write it on issues you delegated."]
    fanout --> join --> coord
  end

  route -->|work it| claim
  route -->|delegate it| fanout
  route -->|record only| park
  live --> done
  coord --> done

  done{"Verified complete?"}
  done -->|yes| close["kata close #60;ref#62; --done<br/>with a message and evidence"]
  done -->|"no, stopping"| park{"Park it?"}
  park -->|start date known| schedule["kata schedule #60;ref#62; #60;date-or-time#62;<br/>sets scheduled_on; clear with -"]
  park -->|no date| someday["kata meta set #60;ref#62; someday true --json-value<br/>clear with kata meta unset #60;ref#62; someday"]
  park -->|no| review["kata label add #60;ref#62; needs-review<br/>plus a comment on what remains"]
```

A user-level Claude Code hook can be installed globally and stay inert outside Kata workspaces:

```json
{
  "hooks": {
    "SessionStart": [
      {
        "matcher": "startup|resume|clear|compact",
        "hooks": [
          {
            "type": "command",
            "command": "if test -f \"$CLAUDE_PROJECT_DIR/.kata.toml\"; then kata quickstart --format contract --workspace \"$CLAUDE_PROJECT_DIR\"; fi"
          }
        ]
      }
    ]
  }
}
```

Codex needs structured JSON rather than plain stdout, so `kata init --with-codex-hooks` covers that case instead. Its contract hook matches `startup|resume|clear|compact`; attention reset stays on `startup|resume|clear`, since resetting attention on compaction would be wrong.

Closes #267

Worth a close look in review: consolidating on one canonical body changes what `--with-agents` writes. The block is now a short prose header plus a DOT flow, rather than the previous prose-and-headings block. The content is stated in kata's own primitives — `needs-review` as a label, `needs-human` as a `work.attention` value, the `work.*` orchestration pair, the fan-out/join coordinator path, and the three parking keys. Existing managed blocks get rewritten in place on the next `kata init --with-agents`, and nothing outside the markers is touched. If you would rather keep the prose form as the committed briefing and render only a subset for injection, say so and I will split them, though that reintroduces the drift this is meant to close.

No schema or migration change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,9 @@ kata close abc4 --done \
 For agent-heavy workspaces, `kata init --with-agents` also writes a managed kata
 briefing into agent guidance files. It refreshes existing real `AGENTS.md` and
 `CLAUDE.md` files, or creates `AGENTS.md` when neither exists, without
-overwriting the rest of either file.
+overwriting the rest of either file. Codex-only workspaces can instead use
+`kata init --with-codex-hooks` to install dynamic contract context and the
+`work.attention` SessionStart lifecycle into `.codex/hooks.json`.
 
 ## Why kata
 
@@ -176,6 +178,47 @@ contract: search before creating, pass an idempotency key on create, prefer
 verified. Close each verified issue promptly with valid evidence and a
 substantive message. [Agent workflows](docs/workflows/agents.md) is the same
 contract in long form.
+
+Agent harnesses can load the shorter managed briefing without changing the
+repository:
+
+```sh
+kata quickstart --format contract
+# equivalent alias; selectors are accepted for user-local integrations
+kata agent-instructions --format contract --workspace /path/to/workspace
+kata quickstart --format contract --project spoke-project
+```
+
+Contract output is marker-free, has no terminal framing, works without an
+initialized workspace, and performs no workspace mutation. It is rendered from
+the same canonical body that `kata init --with-agents` places between its
+managed markers, so static and session-injected guidance cannot drift.
+
+For example, a user-level Claude Code SessionStart hook can remain inert outside
+Kata workspaces while injecting the installed Kata version's contract:
+
+```json
+{
+  "hooks": {
+    "SessionStart": [
+      {
+        "matcher": "startup|resume|clear|compact",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "if test -f \"$CLAUDE_PROJECT_DIR/.kata.toml\"; then kata quickstart --format contract --workspace \"$CLAUDE_PROJECT_DIR\"; fi"
+          }
+        ]
+      }
+    ]
+  }
+}
+```
+
+Codex SessionStart hooks require structured JSON rather than plain contract
+stdout. `kata init --with-codex-hooks` installs Kata's structured adapter for
+startup, resume, clear, and compaction while keeping attention reset limited to
+startup, resume, and clear.
 
 ## Contributing
 
--- a/cmd/kata/agent_contract.go
+++ b/cmd/kata/agent_contract.go
@@ -1,0 +1,68 @@
+package main
+
+// agentsBlockBegin and agentsBlockEnd delimit the section kata manages inside
+// an agent guidance file. They let init refresh kata's guidance in place across
+// re-runs without disturbing anything else a project keeps in that file.
+const (
+	agentsBlockBegin = "<!-- BEGIN KATA (managed by `kata init --with-agents`) -->"
+	agentsBlockEnd   = "<!-- END KATA -->"
+)
+
+// agentContractText is the canonical marker-free briefing shared by dynamic
+// session injection and repository-managed agent guidance.
+const agentContractText = "Kata is the system of record for intent.\n\n" +
+	"- Never `kata delete` or `kata purge` without explicit user authorization.\n\n" +
+	`~~~dot
+digraph kata {
+  rankdir=TB; node [shape=box];
+
+  arrive   [shape=diamond label="Work arrives"];
+  search   [label="Search first; reuse an open issue\nor create one"];
+  route    [shape=diamond label="Work it, or delegate it?"];
+
+  subgraph cluster_work {
+    label="Working a kata-tracked issue";
+    claim  [label="On claim or start, mark it actively tracked:\nkata meta set <ref> work.attention ok\nIn-flight work becomes visible to coordinators\nand dashboards from the moment it is grabbed."];
+    branch [label="If the work happens on a dedicated branch, stamp it once:\nkata meta set <ref> work.branch <branch>\nor bind at creation:\nkata create ... --meta work.branch=<branch> --idempotency-key <key>"];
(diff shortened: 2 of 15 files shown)


salmonumbrella:
@mariusvniekerk hooks are touchy, and I'm weary of injecting context. This is my proposed solution. Up to you and @wesm and @cpcloud.

wesm:
This looks reasonable to me. I'll take it

PR #281, opened 2026-08-19
Add idle shutdown for auto-started daemons

codyw912:
## Summary

Add opt-in idle shutdown for implicitly started, owner-local daemons.

`autostart_idle_timeout` and `KATA_AUTOSTART_IDLE_TIMEOUT` configure the timeout, with a minimum of 10 seconds. The feature is disabled by default and does not affect explicit daemon starts, remote daemons, public listeners, trusted-proxy deployments, or other shared configurations.

This is intended for short-lived CLI and agent sessions where automatic startup is useful, but leaving the daemon resident indefinitely is unnecessary.

## Behavior

- Foreground requests suspend the idle timer and receive a complete new interval when the final request finishes.
- Already-admitted hooks, synchronization passes, reconciliation work, and claim sweeps may drain without extending the foreground deadline.
- Health, ping, and instance probes remain observational and do not renew the timeout. A SIGHUP hook reload does not renew it either.
- Expiration waits for active finite work, while denying unrelated new background work. A hook job refused by idle admission is counted and logged as dropped.
- Foreground activity can safely revive an expiration blocked by drain work without leaving schedulers stuck.
- Scheduled durable work resumes the next time a client starts the daemon.
- An idle exit writes one `kata daemon: idle shutdown after ...` line to the daemon log.
- `kata daemon start` against a running auto-started daemon that advertises `idle_shutdown` stops it and starts an explicit replacement (reported as `replaced` with `replaced_pid`), so an explicit start always yields a resident daemon. A resident daemon is still reported as `already_running`. `kata daemon restart` always produces an explicit daemon.

The effective capability is advertised through the optional `idle_shutdown` health response. `kata mcp serve` uses that daemon-reported state to send marked keepalives for the lifetime of both stdio and streamable-HTTP MCP servers.

## Safety

Idle expiration, explicit stop, listener failure, parent cancellation, and platform stop events converge on one bounded shutdown coordinator. HTTP handlers, background workers, hook jobs, and platform event pumps drain before their dependencies close.

Daemon stop now drains accepted hook jobs instead of dropping them. The dispatcher cancels in-flight hooks two grace windows before the shutdown deadline so a hook slower than the budget is terminated, killed, and reaped while a clean exit is still possible. If work still cannot join within the shutdown budget, the dedicated process exits without closing database or vector resources underneath still-running work. Restart allows additional time for final cleanup and observable process exit.

Continuous GitHub synchronization, federation, timed-claim maintenance, embeddings, or hook processing should continue to use an explicitly managed daemon.

The change includes generated API clients, browser schema updates, operator documentation, and a durable design note.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -1436,6 +1436,8 @@ components:
           $ref: "#/components/schemas/EmbeddingsHealth"
         federation_config:
           $ref: "#/components/schemas/FederationConfigHealth"
+        idle_shutdown:
+          $ref: "#/components/schemas/IdleShutdownHealth"
         ok:
           type: boolean
         schema_version:
@@ -1455,6 +1457,25 @@ components:
         - version
         - uptime
         - started_at
+      type: object
+    IdleShutdownHealth:
+      additionalProperties: true
+      properties:
+        deadline:
+          format: date-time
+          type: string
+        state:
+          enum:
+            - armed
+            - foreground
+            - blocked
+            - stopping
+          type: string
+        timeout:
+          type: string
+      required:
+        - timeout
+        - state
       type: object
     ImportBatchResult:
       additionalProperties: true
@@ -4237,7 +4258,7 @@ components:
       type: object
 info:
   title: kata
-  version: 0.11.0
+  version: 0.12.0
 openapi: 3.1.0
 paths:
   /api/v1/audit/closes:
--- a/cmd/kata/api_compat.go
+++ b/cmd/kata/api_compat.go
@@ -18,6 +18,13 @@ const (
 	apiVersionMCPServer = "0.11.0"
 )
 
+type daemonAPIHealth struct {
+	APISchemaVersion string `json:"api_schema_version"`
+	IdleShutdown     *struct {
+		Timeout string `json:"timeout"`
+	} `json:"idle_shutdown,omitempty"`
+}
+
 // requireDaemonAPIVersion fails closed before a request that uses query
 // parameters older daemons silently ignored. The health endpoint predates the
 // filtered global queries, so it is safe to use as the capability handshake.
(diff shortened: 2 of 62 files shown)


wesm:
Agree with this in principal. I will review the details

wesm:
I squashed the branch to one commit, rebased it onto `main` (resolving conflicts with the 0.15.1 changes), and reviewed it. The review turned up a handful of issues; I pushed a second commit (`76b0f8ba`) that fixes the ones worth fixing in this PR:

- **`kata daemon start` now replaces an idle-eligible auto-started daemon.** Before, it reported "already running" and left the idle timer armed, so hooks/sync would silently stop minutes later. It now checks the live daemon's health; if `idle_shutdown` is advertised it stops that process and starts an explicit one, printing both PIDs (`replaced` / `replaced_pid` in JSON). Resident daemons are still reported as `already_running`.
- **Daemon stop exits cleanly under slow hooks.** `Dispatcher.Shutdown` drained the queue but only cancelled in-flight hooks once the budget expired, so with the default 30s hook timeout and a 25s budget one slow hook meant an unclean exit that skipped closing the store. It now cancels in-flight hooks two grace windows before the deadline so the kill and reap finish inside the budget.
- **Idle exit is logged.** The daemon writes one `kata daemon: idle shutdown after <timeout> without client activity` line to its log, so a "disappeared daemon" is diagnosable.
- **Hook jobs refused by idle admission are counted and logged as dropped**, like a full queue, instead of vanishing silently.
- **SIGHUP hook reload no longer counts as foreground activity** (it was resetting the idle deadline).
- Removed the now-unused foreground-lease adapter, deleted a test assertion that could never fail, fixed a `require` call inside an HTTP handler goroutine, and updated the changelog, config/CLI reference, and design note.

Things I noted but left out of this PR as follow-ups: there is no client retry for the ~ms teardown window (a request can land on `503 daemon_stopping` or a closed socket with a live PID), the MCP keepalive decision is made once at startup, and the HTTP drain timeout and the coordinator budget are two separate constants.

Issue #286, opened 2026-08-21
Support portable external root connectors (Notion, Linear)


salmonumbrella:
Connect Kata to Notion, Linear and any external issue tracker. Portable external-root connectors let a Kata issue stay linked to a root in another system without embedding provider-specific code or credentials in Kata.

## Outcomes

- A versioned executable connector protocol, public Go SDK, and language-neutral conformance contract.
- Durable external-root bindings across SQLite and PostgreSQL, including identity pins, retry claims, conflict state, replay, restore, move/merge, and export/import behavior.
- Reconciliation for externally owned title/body, inbound comments and lifecycle, opt-in outbound comments, mapped planning fields, and verified completion.
- Redacted daemon administration plus connector and bridge CLI workflows.
- Provider-specific adapters remain separate projects.

## Delivery sequence

Review and merge these in order. Each branch is cumulative so it builds before its prerequisites merge; after each earlier PR lands, the next GitHub diff narrows to its listed seam. Every PR contains exactly one squashed commit.

1. #287 — connector protocol, Go SDK, process boundary, author contract, and conformance kit.
2. #291 — SQLite/PostgreSQL storage, PostgreSQL migration 000026, schema version 26, and persistence conformance.
3. #292 — rootbridge reconciliation, retry/identity safety, comments, lifecycle, and planning-field conflict handling.
4. #293 — daemon API, generated client, CLI commands, runner lifecycle, and operator documentation.

## Design constraints

- The external root owns bound title and body. Local edits are rejected with guidance to run `kata bridge unbind <issue-ref>` first.
- Inbound comments and lifecycle are enabled by default. Outbound comments remain opt-in and uncertain delivery is not retried blindly.
- Planning mappings are deliberately limited to `scheduled_on` and `deadline_on`; another field requires an explicit persisted-schema and reconciliation change.
- Connector subprocesses are bounded per RPC in this version. A long-lived transport is deferred.
- No configured connectors means no reconciliation runner or event-wake worker starts.
- Only values explicitly mapped through `connector.env` are treated as secrets and redacted. Arguments and settings remain diagnosable non-secret configuration.
- A web UI binding indicator is out of scope for this series; status is exposed through the API and CLI.

## Current status

- #287 is updated and ready for review.
- #291, #292, and #293 are dependent drafts, submitted in merge order.
[...]

PR #287, opened 2026-08-21
Add connector protocol SDK and conformance kit

salmonumbrella:
## What changed

- define the versioned newline-delimited JSON connector protocol, typed field values, capabilities, structured safe errors, and identity rules
- ship the public Go SDK plus language-neutral conformance transcripts and a reusable conformance suite
- add bounded subprocess invocation, process-tree cleanup, connector auditing, and configuration validation
- redact only secret values explicitly mapped through `connector.env`; command arguments and connector settings remain diagnosable non-secret configuration

## Scope

This is PR 1 of the delivery sequence tracked in #286. It contains only the connector protocol, SDK, runtime boundary, and conformance tooling. Persistence, reconciliation, and daemon/API/CLI integration follow in dependent PRs.

Connector processes are invoked once per bounded RPC in this version. A long-lived transport can be added later without changing the protocol contract.

Part of #286.

### DIFF
--- a/cmd/kata/create_test.go
+++ b/cmd/kata/create_test.go
@@ -204,6 +204,7 @@ func TestCreate_ForceNewBypassesLookalike(t *testing.T) {
 // the actual "broken .kata.toml" the user can fix. The fix-it error
 // must surface client-side without ever calling the daemon.
 func TestResolveProjectID_PropagatesParseError(t *testing.T) {
+	resetFlags(t)
 	dir := t.TempDir()
 	require.NoError(t, os.WriteFile(filepath.Join(dir, ".kata.toml"), //nolint:gosec // test fixture mode matches production
 		[]byte("not = valid = toml ==="), 0o644))
@@ -224,6 +225,7 @@ func TestResolveProjectID_PropagatesParseError(t *testing.T) {
 // through with start_path so the daemon can resolve via its own
 // filesystem walk (local-mode behavior).
 func TestResolveProjectID_FallsBackOnMissingConfig(t *testing.T) {
+	resetFlags(t)
 	dir := t.TempDir() // no .kata.toml
 
 	var got map[string]any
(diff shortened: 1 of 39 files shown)


wesm:
Thanks for this. The feature makes sense for the project and CI is green, including the PostgreSQL conformance runs. Before merging I want to line up on intent, and then there is a short list of fixes.

## Questions on intent

1. **Who owns title and body while an issue is bound?** The PR text says "Kata authoritative," but the code makes the external root the owner: `EditIssue` on a bound issue returns `ErrExternalRootContentOwned`, and each reconcile pass copies the external title/body into Kata. That is a coherent design, but which one did you mean? If "external owns," can you update the description, and make the CLI error point the user at `kata bridge unbind`?
2. **Is it intended that the web UI has no sign an issue is bound?** Today a browser user editing a bound title gets a 409 with no explanation. Is a UI indicator planned as a follow-up, or out of scope?
3. **Should the runner start when no `[[connector]]` is configured?** Right now every daemon starts the 30s scan loop and an event subscriber even with zero connectors. Is there a reason not to skip startup in that case (which would also give operators a kill switch)?
4. **Planning fields are hard-coded to `scheduled_on` and `deadline_on` in the schema CHECK constraints.** Is that deliberate, knowing a third field means a migration?
5. **One subprocess per RPC, about nine per reconcile per binding.** Was a long-lived connector process considered and rejected, or just deferred?

## Fixes needed

- **Dead surface to remove**
  - `EnsureExternalRootReviewLabel` is on `db.Storage`, implemented in both backends, conformance-tested, and never called from production code.
  - `Reconciler` has seven `Run*` entry points that differ only in event-sink plumbing (`internal/rootbridge/reconciler.go:128-250`). One options struct would do.
  - `conformance.Fixture.Handler()` is required and nil-checked but never used by the transcript runner. `validateSafeError` is only called from its own test.
  - `internal/processtree` keeps the old `Prepare`/`TerminateWithGrace` functions alongside the new `Tree` type.
- **Redaction over-reach.** `collectJSONScalars` in `internal/connector/client.go` scrubs every settings scalar, so `verbose = true` in settings turns the word "true" into `[redacted]` in connector error messages. Limit redaction to env-mapped secret values.
- **PR description.** State the persisted schema change explicitly (three new tables, PG migration `000026`) and fix the authority wording per question 1.
- **Commit structure.** One 35K-line commit is hard to review and bisect. Please split along the natural seams: protocol + SDK + conformance, storage + migration, rootbridge, daemon + API + CLI.
- **Docs.** We know these are missing; they can land with the PR or in an immediate follow-up, but `[[connector]]` config, the `kata connector` / `kata bridge` commands, and the connector author contract need coverage.

Things I checked that look good: migration registration, the admin middleware fencing browser and proxy principals, secrets never reaching the DB or API responses, the event-wake skipping the connector's own projection events, and the read-only federated spoke interaction.

PR #291, opened 2026-08-21
Persist external root bridge state

salmonumbrella:
## What changed

- persist external-root bindings, field mappings, field reconciliation baselines/conflicts, retry claims, pending outbound comments, and connector identity pins
- preserve external-root state through export/import, replay, project move/merge, lifecycle, restore, and federation projection rules
- enforce the same storage contract through SQLite and PostgreSQL conformance coverage

## Persisted schema

Both backends add `external_root_bindings`, `external_field_mappings`, and `external_field_states`, with foreign keys, active-binding uniqueness, due-work indexes, typed-field constraints, conflict snapshots, and retry/claim state. PostgreSQL adds migration `000026_external_root_bridges` and advances the schema version to 26; SQLite updates its canonical bootstrap schema to the same logical shape.

## Scope and dependency

This is PR 2 of the delivery sequence tracked in #286 and depends on #287. The branch is cumulative so it builds before #287 merges; after #287 lands, the review diff narrows to storage and migration changes only.

Reconciliation behavior and user-facing daemon/API/CLI surfaces follow in subsequent PRs.

Part of #286.


wesm:
looking now

salmonumbrella:
i paused on one schema suggestion instead of blindly adding it. I want to avoid bloat. 

the two Kata fields are `scheduled_on` and `deadline_on`. they're different:

- `scheduled_on` is a start date and keeps the issue out of ready/next until then
- `deadline_on` is the due date and never gates readiness

for Notion, that would normally be two separate properties, e.g. Start and Due. the edge case is that today both Kata fields can be mapped to the same Notion property, so reads/writes collapse into one value.

the suggested DB constraint would only prevent that within one connector instance: `(connector_instance, external_field_id)` stays unique while active. i haven't added it yet because i want to confirm the product call first: keep both mappings and require distinct external properties, or simplify to one date mapping? wdyt

PR #292, opened 2026-08-21
Reconcile external root bridges

salmonumbrella:
## What changed

- reconcile external-root title, body, comments, lifecycle, and mapped planning fields through one `Reconciler.Run` entry point with explicit run options
- pin connector/account identity, preserve retry claims, recover ambiguous outbound comment publication by replaying stable operation IDs without depending on provider timestamp precision, and verify external completion
- support inbound comments and lifecycle by default while keeping outbound comment publication opt-in
- keep planning reconciliation deliberately limited to `scheduled_on` and `deadline_on`

## Intent decisions

The bound external root owns title and body. Kata imports those values and rejects local edits with a `kata bridge unbind <issue-ref>` recovery hint. A third planning field is not accepted implicitly because it requires an explicit storage migration and reconciliation contract.

The current implementation performs bounded subprocess RPCs per reconciliation operation. A long-lived connector transport is deferred.

## Scope and dependency

This is PR 3 of the delivery sequence tracked in #286 and depends on #287 and #291. The branch is cumulative so it builds before those PRs merge; after they land, the review diff narrows to `internal/rootbridge`.

Daemon/API/CLI integration follows in the final PR. A web UI binding indicator is out of scope for this series.

Part of #286.

### DIFF
--- a/.roborev.toml
+++ b/.roborev.toml
@@ -3,6 +3,16 @@ display_name = "kata"
 review_guidelines = """
 Latest stable Go. Be pragmatic — theoretical findings produce churn.
 
+## Stacked delivery boundaries
+
+Kata may land a cumulative delivery sequence as separate protocol, storage,
+reconciliation, and runtime-integration pull requests. Review each pull request
+against the scope and dependency boundary stated in its description. Missing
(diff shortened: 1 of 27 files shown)


wesm:
rebasing now

PR #296, opened 2026-08-24
refactor(db): simplify import and export representations

salmonumbrella:
## Summary

- Replace the 22-pointer import union with typed payload records while keeping malformed replay rejection before either backend opens a transaction.
- Give each orphan row one explicit drop-or-scrub disposition with arrival-order-independent precedence.
- Centralize event orphan handling across all five export version bands and delegate issue-sync status export to the existing shared path.

## Compatibility

- Preserve NDJSON kind strings and persisted schemas.
- Preserve the per-version event export SQL, including the legacy V1 shape.


wesm:
rebasing

PR #300, opened 2026-08-24
Simplify storage reads and federation rebuilds

salmonumbrella:
This simplifies several internal storage paths while preserving existing API behavior. PostgreSQL now uses shared adapters to read and write timestamps and integer-backed booleans consistently in the claim, sync, federation, project, recurrence, token, and issue paths touched here. Both database backends use the same helpers to order and filter folded event state, and PostgreSQL reuses each project's event stream during a federation rebuild instead of reading the current project twice.

Issue hydration now gets parents, children, blockers, and related issues through one structured storage call while keeping each relationship's existing visibility rules. Federation enrollment listing and revocation now query within the requested project at the database boundary instead of loading instance-wide history and filtering it in service code.

This does not change persisted schemas or migrations. The instance-wide enrollment audit remains global.


wesm:
looking

PR #302, opened 2026-08-25
Consolidate federation replica leave state and enrollment cleanup

salmonumbrella:
Federation replica leave tracking was spread across three process-global maps (leave intents, suppression, and in-flight hub operations) plus a second drain counter duplicated in the rebind path, and the mid-flight enrollment compensation logic existed twice: inline in the daemon's hub-operation finish closure and again in the config reconciler.

This change replaces the maps with one keyed registry (`internal/daemon/federation_replica_transitions.go`) that owns each mapping's leave state (idle, leave-pending, left) and the drain accounting for hub operations already in flight. The registry keeps its own lock only so the config reconciler can read suppression without blocking on `ensureFederationReplicaMu`, which is held across database and credential-store I/O; every state mutation still happens under that mutex, preserving the existing lock ordering, operator error messages, and rebind behavior. A leave becomes observable to those lock-free reads only after its durable credential mark succeeds, so a failed preparation never parks the reconciler as reconciled and never erases a completed leave's suppression.

Both windows where an explicit leave can land mid-enrollment — before the hub operation finishes, or while local convergence runs — now go through one durable pending-enrollment stamp (`daemon.RecordFederationReplicaPendingEnrollment`) and one `compensateEnrollment` policy. An observed leave (a leave-pending reservation stamp or in-process leave state) wins over an enrollment or convergence error, and hub cleanup failures are tolerated only while a leave is prepared or completed in this process, because the durable stamp lets the leave's own cleanup revoke the enrollment later. When no leave explains a reservation change during convergence, the conflict keeps failing — after revoking any enrollment the operation created — because the controller stops retrying a mapping whose reconciliation reports success. Previously a successful revocation reported the unconverged mapping as reconciled.

No route, wire-format, or persisted-schema changes. The only new exported symbol is `daemon.RecordFederationReplicaPendingEnrollment`, used by the config reconciler.

### DIFF
--- a/internal/daemon/federation_rebind_service.go
+++ b/internal/daemon/federation_rebind_service.go
@@ -9,7 +9,6 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
-	"sync"
 	"time"
 
 	"go.kenn.io/kata/internal/api"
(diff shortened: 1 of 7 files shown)


wesm:
looking at this, will rewrite PR title and description

PR #304, opened 2026-08-25
Preserve daemon credential provenance

salmonumbrella:
## What changed

- carry daemon source, credentials, transport policy, and Unix endpoint metadata through resolution instead of reconstructing them from URLs
- centralize bearer-target validation and origin-pinned client construction
- route CLI and TUI callers through resolved daemon clients while keeping hub credentials isolated from local daemon credentials
- preserve non-starting health probes, long-running GitHub sync requests, and the existing federation redirect boundary

## Why

Reconstructing a daemon client from its URL throws away where the endpoint came from and which credential and transport policy belong to it. That makes unsafe plaintext failures late and opaque, and it creates too many places where a local daemon token could be paired with the wrong origin.

## Usage

Configuration syntax is unchanged. Plaintext targets without the required private-network trust or `allow_insecure` opt-in now fail when the client is built.

### DIFF
--- a/cmd/kata/client.go
+++ b/cmd/kata/client.go
@@ -1,13 +1,17 @@
 package main
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
 	"os"
+	"strings"
 	"time"
 
+	"github.com/spf13/cobra"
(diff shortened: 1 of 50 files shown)


wesm:
rebasing

PR #305, opened 2026-08-25
Simplify CLI wait and output state

salmonumbrella:
## What changed

- represent `kata wait` target state, join mode, and deadlines as single typed values
- resolve CLI output mode once while preserving legacy import and contract-conflict behavior
- replace overlapping error-path argument walkers with one total pre-parser

## Why

Wait state and output selection were spread across overlapping booleans and parsers that could disagree on the same command line. The CLI should make each decision once and carry that answer through normal execution and error reporting.

## Usage

No usage change.


wesm:
rebasing

PR #306, opened 2026-08-25
Replace positional TUI state with typed identities

salmonumbrella:
## What changed

- address centered-form fields by stable identity instead of positional indexes
- derive hub-project labels, routing, selection, and counts from one typed row
- model enroll and leave as one tagged federation operation with a shared monotonic attempt counter
- discard superseded federation results when the daemon or active flow changes

## Why

Positional state makes unrelated TUI fields and asynchronous federation flows easy to cross-wire after a view changes. Stable identities and tagged operations keep state attached to the thing it represents, including during daemon switches and delayed responses.

## Usage

No usage change.


wesm:
rebasing

PR #308, opened 2026-08-25
Make JSON schemas and lease aliases self-describing

salmonumbrella:
## What changed

- let opaque JSON storage and API types define their own OpenAPI shapes
- remove the daemon's property-name schema rewrite while keeping committed OpenAPI documents and generated clients byte-identical
- mirror canonical `lease` fields to deprecated `claim` aliases once at response boundaries
- keep explicit `claim` fallbacks where first-party clients still talk to older hubs
- leave persisted schemas, migrations, and wire spellings unchanged

## Why

Schema meaning and lease compatibility were encoded in property names and repeated producer logic. That is fragile: an unrelated typed field named `metadata` could be rewritten, and lease aliases could drift between response paths.

## Usage

No usage change.

### DIFF
--- a/cmd/kata/federation_lease.go
+++ b/cmd/kata/federation_lease.go
@@ -301,15 +301,31 @@ func explicitClaimAdminReason(cmd *cobra.Command, reason string) (string, error)
 	return strings.TrimSpace(reason), nil
 }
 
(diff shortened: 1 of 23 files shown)


wesm:
rebasing

PR #311, opened 2026-08-26
Standardize web actions on Kit UI

### LINKED ISSUE #288: In Web UI Dark theme some buttons still use white background with light color font
Image

salmonumbrella:
## What changed

- move the remaining generic web actions onto Kit UI buttons while preserving the specialized navigation, menu, and graph controls
- use Kit UI theme state, semantic tokens, screen-reader utilities, and the standard responsive breakpoint across the web shell
- make the normal web check reject raw generic controls that drift outside Kit UI
- cover dark read-only action contrast and theme-listener cleanup through browser and component regressions

## Why

Generic actions still mixed hand-rolled button styles with Kit UI, so disabled controls could drift to a light background and low-contrast text in dark mode. Keeping those actions and theme state on the shared primitives makes the dark theme consistent and prevents the same regression from returning.

## Usage

No usage change.

Closes #288


salmonumbrella:
cc: @mariusvniekerk

mariusvniekerk:
NICE

PR #330, opened 2026-09-02
fix(ownership): add compare-and-swap guards

### LINKED ISSUE #328: claim: distinguish a same-actor re-claim; unassign with an expected owner
`kata claim` is actor-scoped. When an actor claims an issue it already owns, the call exits 0 with `changed:false` and no error. Two agent sessions on one machine share `KATA_AUTHOR`, so to kata they are one actor: the second session's claim is a silent no-op and both sessions proceed on the same issue. That is how two of my sessions ran the same fleet migration concurrently on 2026-08-03.

I have since built a per-session marker on top of `claim` in my own tooling (a comment written after the claim, read back before the next one). It works, but a comment is not a lock: every window between two kata calls stays open, and the reviews keep finding new interleavings in those windows.

Two small additions would let a client close this properly:

1. **A way to tell a same-actor re-claim from a sibling's.** Either `claim --if-unowned` (non-zero when the issue is already held, whoever holds it), or `claim --session <token>` that is recorded with the claim and returned on a later `changed:false`, so a client can distinguish "my own retry" from "another session under my actor name". Today `changed:false` is the only signal and it is identical for both.

2. **`unassign --expect-owner <actor>`** as a compare-and-swap: release only if the owner is still the one the caller read. The window between an owner read and the unassign is the residual risk my release paths document today.

Repro on main (806ef952):

```
export KATA_AUTHOR=joi-mac
kata claim proj#abc --json   # {"changed":true,...}
kata claim proj#abc --json   # {"changed":false,...}  exit 0, no error
```

salmonumbrella:
## What changed

- Added `kata claim --if-unowned` for an atomic claim that conflicts whenever an issue already has an owner, including the same actor.
- Added `kata unassign --expect-owner <owner>` for an atomic compare-and-swap release.
- Implemented both guards in SQLite and PostgreSQL, exposed them through the API and generated clients, and kept the existing unguarded behavior compatible.

## Why

Separate sessions can share one actor name. A same-actor re-claim therefore cannot tell a retry from a sibling session taking the same work, and a read followed by an unconditional unassign leaves a race with reassignment. These guards let clients close both windows in one storage transaction.

## Usage

```sh
kata claim spoke-project#abc4 --if-unowned
kata unassign spoke-project#abc4 --expect-owner worker-a
```

`--if-unowned` is mutually exclusive with `--force`. An expected-owner mismatch returns a conflict without changing ownership.

Closes #328

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -230,6 +230,8 @@ components:
           type: string
         force:
           type: boolean
+        if_unowned:
+          type: boolean
       required:
         - actor
       type: object
@@ -4587,6 +4589,8 @@ components:
       additionalProperties: false
       properties:
         actor:
+          type: string
+        expected_owner:
           type: string
       type: object
 info:
--- a/cmd/kata/assign.go
+++ b/cmd/kata/assign.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"strings"
 
 	"github.com/spf13/cobra"
 )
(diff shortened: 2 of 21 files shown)


salmonumbrella:
@wesm ready

wesm:
Cool, looks good

PR #331, opened 2026-09-02
fix(create): bound look-alike checks and clarify outcomes

### LINKED ISSUE #329: kata create with a ~12 KB body times out in the look-alike check; --force-new returns in under a second
Client: kata v0.15.1-25-g2bc1a707 on macOS, talking to the daemon over a tailnet (plain HTTP, ~100 ms RTT). Project has about 3,400 issues.

What happened: `kata create "<title>" --body-file body.md --label ... --related x6 --meta ... --idempotency-key ...` with an 11.8 KB body failed twice in a row with

```
kata: Post "http://.../api/v1/projects/15/issues": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```

The same command with `--force-new` added returned the created issue in 0.86 s. Reads (`kata search`, `kata show`, `kata comment`) were fast the whole time, so it is not the network or the daemon being down.

So the look-alike soft-block check looks like the slow part, and it scales badly enough with body length or project size to blow the client timeout. Two things would help:

1. Run the look-alike comparison on the title plus the first N bytes of the body, or cap its time budget and skip the check with a warning when it runs out, instead of failing the create.
2. When the client times out on create, say so in the error: something like "look-alike check timed out; retry with --force-new" would have saved the second attempt.

Happy to try a patched build against this daemon if useful.

salmonumbrella:
## What changed

- Bounded look-alike retrieval and scoring to the first 500 Unicode code points of both the title and body.
- Classified ambiguous create transport failures, including timeouts, cancellations, and dropped responses, as `create_outcome_unknown`.
- Added recovery guidance for normal and `--force-new` creates without exposing the daemon URL.

## Why

Large issue text could expand into expensive look-alike queries even though text after the scorer's 500-code-point boundary could not affect the result. If the connection failed after sending a create, the raw error also left users without a safe way to decide whether retrying would create a duplicate.

## Usage

No command changes are required. If `kata create` reports `create_outcome_unknown`, check whether the issue exists before retrying. Use `--force-new` only after confirming that no issue was created.

Closes #329

### DIFF
--- a/cmd/kata/create.go
+++ b/cmd/kata/create.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"os"
 	"strings"
@@ -172,7 +173,7 @@ func newCreateCmd() *cobra.Command {
 			fmt.Sprintf("%s/api/v1/projects/%d/issues", baseURL, projectID),
 			headers, req)
 		if err != nil {
-			return err
+			return createRequestError(err, forceNew)
 		}
 		if status >= 400 {
 			return apiErrFromBody(status, bs)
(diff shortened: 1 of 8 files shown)


wesm:
looking

PR #332, opened 2026-09-02
Make issue close safe to retry with an idempotency key

naveenspark:
Closing an issue can now be retried safely after a lost response. Send `Idempotency-Key` on the close request (`kata close --idempotency-key`, MCP `idempotency_key`) and a retry returns the original close instead of running a second one. `If-Match` (`kata close --if-match`, MCP `revision`) rejects a close of an issue that changed underneath you.

Before this, a client that timed out after the daemon committed had no safe move. Retrying gave a no-op with no event, and the CLI's follow-up `--comment` could post twice.

What a retry gets:

- The same key with the same request returns the original receipt. A different body, actor, issue, or revision guard gets 409.
- If someone else closed the issue first, or moved it before the write, the keyed request gets 409 rather than a silent no-op.
- If the database committed but the response was lost, the daemon publishes the committed events once, then replies with the receipt.
- A replayed receipt follows the same visibility rules as a fresh request: active project, correct route, and the caller's host scope.
- Retry headers require `retry_protocol: "close-v1"` in the body. Old daemons reject the unknown field instead of silently ignoring the headers. API schema version is 0.15.0.

No schema change. Hook delivery stays in memory, so a daemon crash between commit and publish loses hook jobs for a close the same way it does for every other mutation.

Also: a comment idempotency key now belongs to one issue, not one project, and a comment retry survives a project move.


wesm:
looking

wesm:
Accepted as is

PR #333, opened 2026-09-02
Add revision and body excerpts to agent-mode search and list rows

naveenspark:
Agents reading `kata search --agent` now get enough context to choose a result without running `kata show` on every hit. Each search row appends `owner`, `priority`, `revision`, and an `excerpt` from the issue body. `kata list --agent` rows append `revision`, and `kata show --agent` prints a `Revision:` line.

Before this change, a search row carried only the id, score, status, matched fields, and title. An agent had to open each candidate to learn who owned it, how urgent it was, or what the body said. No agent-mode read exposed the revision that `--if-match` writes need, so agents had to fall back to `--json` for it.

The excerpt is at most 160 characters. It takes up to 24 words of the body around the first word that matches a query term, with `…` marking trimmed ends. If that window is still over budget, it is re-centered on the matched term so long lead-in text cannot push the match out of view. Term matching folds case and diacritics and splits on the same letter/number boundaries the lexical search backends use, so `cafe` finds `café`, `foo-bar` finds `foo bar`, and `log` does not match inside `catalog`. When no term matches, such as a title-only or semantic hit, the excerpt is the start of the body. Empty bodies emit no `excerpt=` field.

All new fields are appended after the existing ones, so `agent_format` stays `1` and current parsers keep working. `docs/reference/agent-output.md` documents the new row shapes.

### DIFF
--- a/cmd/kata/list.go
+++ b/cmd/kata/list.go
@@ -137,6 +137,7 @@ func newListCmd() *cobra.Command {
 					Status      string   `json:"status"`
 					Owner       *string  `json:"owner"`
 					Priority    *int64   `json:"priority"`
+					Revision    int64    `json:"revision"`
 					Labels      []string `json:"labels"`
 					Blocked     bool     `json:"blocked"`
 					Parent      *struct {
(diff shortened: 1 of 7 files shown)


wesm:
looking

PR #334, opened 2026-09-02
Add `kata federation lease renew` to extend a timed lease before it expires

naveenspark:
Timed federation write leases could be acquired from the CLI with `kata federation lease acquire --ttl`, but there was no CLI way to extend one. A holder whose lease was about to expire had to release and re-acquire, which drops the lease identity and opens a window for another actor to take the issue. The daemon and the MCP lease tool already supported renewal; only the CLI lacked it.

`kata federation lease renew <issue-ref> --ttl <duration>` now posts to the daemon's existing renew action. The lease keeps its claim UID, its revision advances, and the new expiry is computed from the hub's clock, so the holder keeps exclusive coordination without a gap. `--ttl` is required and accepts the same whole-unit 60s–24h durations as acquire.

Renewal applies only to timed leases. Renewing a hard lease (one acquired without `--ttl`) fails with the daemon's `hard claims cannot be renewed` validation error, and renewing a lease held by someone else or no lease at all fails with `claim not held`. Both surface as ordinary CLI errors with non-zero exit; there is no path that reports success without the hub extending the lease.

Human output prints `renewed lease on <ref> as <holder>`, agent output reports `State: renewed`, and JSON output is the daemon's lease response unchanged. The command is documented in the CLI reference and the federation design doc.

### DIFF
--- a/cmd/kata/federation_lease.go
+++ b/cmd/kata/federation_lease.go
@@ -55,12 +55,12 @@ func newFederationLeaseRenewCmd() *cobra.Command {
 		Short: "renew a timed federation write lease",
 		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if strings.TrimSpace(ttlRaw) == "" {
-				return &cliError{Message: "--ttl is required", Kind: kindUsage, ExitCode: ExitUsage}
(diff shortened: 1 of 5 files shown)


wesm:
looking

PR #337, opened 2026-09-03
fix(create): handle remaining ambiguous outcomes

salmonumbrella:
## What changed

- Bounded look-alike retrieval and scoring to the first 500 Unicode code points of both the title and body.
- Classified dropped connections and truncated responses as `create_outcome_unknown`, including platform-specific Unix and Windows failures.
- Updated the CLI reference to describe the final matching and recovery behavior.
- Gave the hook-dispatcher drain test enough time for sequential process starts on Windows runners.

## Why

Follow-up review of #331 found additional paths that could still produce oversized look-alike queries or leave a successful create looking like a raw transport failure. Those cases need the same check-before-retry guidance so callers do not accidentally create duplicates.

## Usage

No command changes are required. If `kata create` reports `create_outcome_unknown`, check whether the issue exists before retrying. Use `--force-new` only after confirming that no issue was created.

Follow-up to #331.

### DIFF
--- a/cmd/kata/conn_dropped_unix.go
+++ b/cmd/kata/conn_dropped_unix.go
@@ -1,0 +1,12 @@
+//go:build !windows
+
+package main
+
+import (
+	"errors"
+	"syscall"
+)
+
+func connDroppedErrno(err error) bool {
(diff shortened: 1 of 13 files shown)


salmonumbrella:
@wesm ready

Issue #345, opened 2026-09-06
Wrong styles in dark mode for top right buttons


toqueteos:
<img width="295" height="175" alt="Image" src="https://github.com/user-attachments/assets/48b4c72d-1cda-49ca-a95a-a817ec3fa96a" />

It seems to be [this](https://github.com/kenn-io/kata/blob/905ff9b76ad3c10ac03cec8b2ded56790e0e4abe/packages/kata-ui/src/IssueDetail.svelte#L27-L36) `div.host-actions`

It was like this in previous versions and latest `v0.16.0` is still affected

wesm:
looking

wesm:
what browser is this in?

toqueteos:
> what browser is this in?

@wesm My bad, forgot to report system details

That screenshot is taken on Vivaldi (a Chromium fork) in Linux (Bazzite) using an AMD 7700X CPU.

## Vivaldi

```
Version: 8.2.4133.47 (Official Build) (x86_64)
Chromium Version: 152.0.7977.112
Channel: Official Build
Platform / OS: Linux - org.freedesktop.platform 25.08
Executable Path: /app/vivaldi/vivaldi
Command Line:  /app/vivaldi/vivaldi --enable-features=WebRTCPipeWireCapturer --disable-features=WebAssemblyTrapHandler,DesktopPWAsRunOnOsLogin --class=Vivaldi-flatpak --no-default-browser-check --ozone-platform=wayland --flag-switches-begin --flag-switches-end --render-node-override=/dev/dri/renderD128
```

## Chrome

I can also confirm same behavior on Chrome (`Version 152.0.7977.82 (Official Build) (x86_64)`) using same system Linux/AMD.

## Safari

I can also reproduce the issue on Safari (`Safari 26.6 (21624.4.5.11.5`) on Mac Mini M4 (`Tahoe 26.6.1`).

## Firefox

On Firefox (using same system Linux/AMD) I can't get past the login, it's stuck in:

<img width="385" height="105" alt="Image" src="https://github.com/user-attachments/assets/61330a2a-e095-40ef-a4a4-54f0b25537b1" />

But that seems like another entire thing which I'm going to move past over for now.

Probably an edge case of #285

wesm:
[...]

PR #350, opened 2026-09-08
Add daemon restart reminder to `kata update`

toqueteos:
Might help avoid some tickets in the future (see #345)


wesm:
closing in favor of #352, this should not be something you have to worry about

PR #358, opened 2026-09-11
Stack the issue editor heading when the pane is narrow

ruaridhw:
Editing an issue on a narrow pane squashes the title: `.detail-heading` is a row, `.detail-actions` is `flex: 0 0 auto` with a `nowrap` timestamp, so `.detail-heading-main` absorbs the whole squeeze and the title wraps a word at a time.

Stacking the title above the actions below 640px fixes it. Keyed to the pane, not the viewport, because the detail pane is a resizable `SplitLayout` secondary (`minSecondary: 220`) and can be dragged narrow while the window is still wide — a `@media` query never fires there. Same reasoning as the `@container list` rules in `IssueCollection.svelte`.

The read-only view already handles this (`packages/kata-ui/src/IssueDetail.svelte` stacks `.detail-header` at 640px); the editor had no width handling at all.

No change above 640px.

### Current

before

### With change

after

### DIFF
--- a/web/src/components/IssueEditor.svelte
+++ b/web/src/components/IssueEditor.svelte
@@ -558,6 +558,8 @@
     overflow: auto;
     background: var(--bg-primary);
     padding: 18px 22px;
+    container-type: inline-size;
+    container-name: kata-detail;
   }
 
   .mutation-controls {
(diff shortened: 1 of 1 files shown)


mariusvniekerk (approved):

salmonumbrella:
sick thanks @ruaridhw

Issue #355, opened 2026-09-10
[feat idea] Taggable team members + auto-inject into Pi/Harness (details inside)


taylorsatula:
I love Kata but because it is agent based I have found it hard to tag team members when I need them for a Ticket without sending a Slack message. What if there was a way for Kata to inject into a root Agents.md or something autoloaded a "Alert the user that Taylor needs their eyes on ticket k18d when you have a moment. Surface this alert to the user when contextually appropriate.".  Then when its closed out it is removed from the AGENTS.md (or whatever) so that it is cleared when the next Pi session loads.

Kind of like a heads-up display.

Thoughts?

salmonumbrella:
@wesm this can be closed. thanks for the idea @taylorsatula

taylorsatula:
:D any time. Thanks for the quick turnaround.

salmonumbrella:
@taylorsatula can you please close this issue?

PR #359, opened 2026-09-11
Add teammate attention requests and a project inbox

salmonumbrella:
## What changed

- Adds `kata notify` to request a teammate's attention on an issue and clear the request after it is handled.
- Adds a project-scoped `kata inbox`, including bounded context for external harness integrations. Closed issues leave the inbox; reopening restores uncleared requests.
- Explains teammate attention requests in the shared DOT workflow and CLI quickstart.
- Uses existing issue metadata with per-recipient patches. Messages are limited to 1024 bytes. No new backend API, schema, dependencies, or harness code.

## Why

Agents collaborating on a shared goal need a way to ask another teammate to pick something up. This gives Kata a small request-and-read workflow that an external harness can consume for review, investigation, or a handoff.

## Usage

```sh
kata notify abc4 --to teammate --message "Please check the reproduction"
kata inbox --for teammate
# After handling the request:
kata notify abc4 --to teammate --clear

export KATA_INBOX_USER=teammate
kata inbox --context --workspace /path/to/workspace
```

`teammate` is an example recipient handle. Use the same exact handle for notify, inbox, and clear.

The harness adapter belongs outside Kata. It should fetch context before a prompt, inject successful stdout transiently, and drop the previous context on empty output or failure. Kata does not install a harness adapter or write live alerts into `AGENTS.md`.

Refs #355


salmonumbrella:
@wesm ready

wesm:
rejected

PR #365, opened 2026-09-14
fix(mcp): avoid top-level oneOf/allOf in tool input schemas

hsb3:
The Anthropic Messages API rejects a tool `input_schema` with `oneOf`, `allOf`, or `anyOf` at the top level (`input_schema does not support oneOf, allOf, or anyOf at the top level`). Nine MCP tools publish such schemas: `kata.close`, `kata.edit`, `kata.list`, `kata.ready`, `kata.set_deadline`, `kata.set_schedule`, `kata.recurrence_update`, `kata.federation_leave`, and `kata.bridge_resolve_comment`. A client that passes MCP tool schemas through to that API gets a 400 for any request that includes one of them, so it has to drop the tools or rewrite their schemas itself.

This keeps the same validation rules and expresses them with keywords that are accepted at the top level:

- Field exclusions (`list`, `ready`, `edit`) become a root `not`, with the exclusions nested in an `anyOf` inside it when a tool has more than one.
- The single conditional rule on `recurrence_update`, `federation_leave`, and `bridge_resolve_comment` moves from a one-element `allOf` to root `if`/`then`/`else`.
- The exactly-one-of rule on `set_deadline` and `set_schedule` becomes `if`/`then`/`else`.
- The per-reason evidence rules on `close` become an `if`/`then`/`else` chain keyed on `reason`. The last reason's rules are the final `else`, so a missing or unknown reason still fails validation.

The tool catalog test now asserts that no tool publishes top-level `oneOf`, `allOf`, or `anyOf`, and a new schema test covers the `edit` and `list` field exclusions.


wesm:
reviewing

PR #368, opened 2026-09-14
Fix flaky Windows readiness tests

rodboev:
On Windows, connector readiness polling now retries `ERROR_SHARING_VIOLATION` during the readiness-file handoff and still fails on other unexpected errors.

The daemon connector-status test allows ten seconds for startup, including fresh database initialization, and reports runtime-file read or JSON decode errors on timeout. It continues to require HTTP 200 with unhealthy connector status and a redacted command path.


wesm:
looking

rodboev:
Oops, didn't realize this one was left red. Sorry!

PR #377, opened 2026-09-15
feat(notify): surface reached planning dates in inbox

salmonumbrella:
Reached `scheduled_on` and `deadline_on` values now surface through the existing `notify.<actor>` inbox metadata. A daemon sweep delivers one notification to the owner or author, preserves manual notifications, deduplicates acknowledged notices through event history, and prevents enabled federation spokes from originating duplicates.

This also documents the inbox behavior and covers SQLite/PostgreSQL parity.


wesm:
rebasing

PR #398, opened 2026-09-18
feat: add expiring issue assignments

### LINKED ISSUE #396: [feat idea] Claim tickets with a timeout so two people don't try to implement the same thing twice.
I hope its okay that I keep sending these as I encounter stuff. I've been working with Kata every day and stuff pops up as I use it. Don't feel bad declining the ones that are dumb or scope creep.

I've experienced situations where I'll get started on a kata ticket and then someone else on the team also picks it off the tree then theres a race where neither of us know that someone else is actively working on implementing the ticket. Propose: the ability to get a temporary "lock" of sorts on a ticket so that when another teammate's agent goes to pull a ticket they see "Oh, XYZ Person is already working on this. Check with them via slack or whatever".

salmonumbrella:
## What changed

- `kata claim <ref> --ttl <duration>` creates a temporary assignment for one minute through 24 hours. Repeating the claim renews it, while omitting `--ttl` keeps the existing permanent assignment behavior.
- HTTP and MCP use `ttl_seconds` for the requested lifetime. The resulting expiry remains `assignment_expires_on` everywhere it is displayed or persisted.
- `ready` and `next` treat an expired assignment as unassigned immediately. Background cleanup records the expiry without clearing a concurrent renewal or takeover, including when Kata runs as an embedded service.
- Assignment expiry is visible across SQLite, PostgreSQL, JSONL, federation, Web, and TUI through `assignment_expires_on` and explicit renewal and expiry events.
- Federated temporary assignments remain hub-authoritative. Enrollment claim credentials must provide a TTL and cannot create permanent assignments directly.

## Why

Two agents can currently start the same issue because a permanent assignment either stays around until someone clears it or is avoided entirely. Temporary assignments make active work visible to other agents, then return abandoned work to the queue automatically without adding a separate lock or claim identity.

## Usage

```sh
kata next --unowned --agent
kata claim abc4 --if-unowned --ttl 30m --agent

# Renew before it expires.
kata claim abc4 --ttl 30m --agent
```

`--ttl` accepts whole-second Go durations from `1m` through `24h`. Omit it to create or preserve a permanent assignment.

Closes #396


wesm:
Updated and rebased this PR onto current `main`, preserving the original author.

- Fixed timed-claim renewal after a hub refreshes its snapshots. A renewal could succeed while the spoke still showed no owner; the response now includes the assignment history needed to update the spoke.
- Closing an issue now cancels its assignment timer and keeps its owner. Reopening does not restart the timer, so completed work no longer loses ownership later.
- Kept the existing `already_claimed` conflict code so scripts do not need to change.
- Clarified that people sharing one spoke enrollment share its claim identity, and documented `kata assign` as the way to make a temporary assignment permanent.
- Preserved the revision checks added on `main`. Assignment changes, including renewal and expiry, advance the revision so a later `--if-match` write detects them.

One correction to the earlier assessment: normal SQLite startup already performs the JSONL cutover automatically; operators do not need to run it manually.

Local Go checks passed for the CLI, end-to-end flows, daemon, SQLite/storage, JSONL, MCP, TUI, client, and service packages. Lint and documentation checks also passed. PostgreSQL database tests were skipped locally because Docker access is denied.

wesm:
Fixed the missing assignment expiry in the SQLite cutover exporter. Its JSONL output now preserves the timer on v29 databases and still reads older databases without that column.

The impact is narrower than the review states: normal v29 exports already preserve expiry, and startup leaves v29 databases untouched. This fixes the exporter used when those databases need a later schema upgrade.

A direct export/import test reproduced the lost timer before the fix and passes now. Coverage includes v28 permanent assignments and v29 permanent and timed assignments. The JSONL and database-opening tests and lint passed.

PR #405, opened 2026-09-19
feat(tui): undo recent issue actions

### LINKED ISSUE #404: [feat] have action [u]ndo in tui
I accidentally type stuff in the tui meant for a different terminal. Wreaks havoc. (reopens tickets, closes tickets, parents tickets, etc.) A [u] button to undo tui actions would be swell.

salmonumbrella:
## What changed

- Added `u` in the TUI issue list and detail view to undo the latest eligible issue action. It keeps up to 20 actions in the current session and follows the affected issue across navigation and projects.
- Supports close, reopen, owner, priority, label, body, and newly added parent, blocker, and related links. Undo creates an attributed inverse action while preserving the original audit event.
- Checks the daemon, principal, issue identity, revision, affected field, and exact link before writing. Conflicts leave the current issue untouched. Actions that cannot be safely reversed clear the undo history, including timed assignments and issue or comment creation.
- Updated the TUI help and CLI reference. No database schema change.

## Why

Typing into the TUI when another terminal has focus can close, reopen, or reparent an issue by accident. Recovering now requires manually finding and reversing each change, sometimes after navigating away from the issue.

## Usage

Press `u` in the issue list or detail view to undo the latest eligible action. Press it again to walk backward through the session history. Undoing a reopen opens the normal completion form when the daemon requires evidence.

Closes #404


wesm:
Updated the PR to fix four user-visible problems:

- A conflicting undo now skips that entry and explains why. The next `u` reaches earlier history, including after a conflict in the completion form.
- Discarding an outdated list response now starts a fresh request. The list can finish loading without waiting for a server event.
- Quick successive edits wait for earlier writes instead of failing immediately. Waiting edits keep their arrival order and still respect the existing request timeout.
- Normal edits reuse the connection's daemon identity and permissions, removing one request per edit. Undo still checks them against the live daemon.

I kept the fresh issue read because the detail pane may be stale. I also kept the revision check: a real-daemon test confirms that it catches an issue being reopened and closed again by another client, even when the status and close reason match. Field equality alone misses that change.

The docs now explain skipped conflicts and the lack of redo. If you undo by mistake, repeat the action manually. I left the cosmetic refactor and rare HTTP 408 case out of this patch.

The TUI suite passes with the race detector; the Go build, lint, and documentation checks also pass. The full suite hit four failures outside these changes: `TestInit_RemoteClient_SendsNameNotPath` (passes alone), `TestMakeInstallBakesGitDescribeVersion` and `TestMakeBuildBakesGitDescribeVersion` (Vite missing), and `TestRenderHomebrewCoreFormula_ProducesValidRuby` (Ruby version unset).

wesm:
documented limitation

PR #407, opened 2026-09-19
feat(tui): add Emacs navigation shortcuts

### LINKED ISSUE #397: Support Emacs keybindings for TUI navigation
I think adding simple shortcuts like C-v for forward paging, M-v for backward, C-M-< and C-M-+> for navigating to the start/end, respectively would make for nice additions to the TUI

salmonumbrella:
## What changed

- Added `C-v` and `M-v` as page forward and back shortcuts in the issue list and detail view.
- Added `C-M-<` and `C-M->` for the first and last list item or detail document page. `g` and `G` work in detail too.
- Added `C-n` and `C-p` for down/up movement, and `C-j` and `C-k` to cycle detail activity tabs and the child section.
- Updated TUI help and CLI docs. Help in split layout keeps the panes still while it is open.

## Why

The TUI already has page and boundary navigation, but those actions require different keys from Emacs. These shortcuts let people use familiar keys without changing the existing ones.

## Usage

Press `C-n`/`C-p` to move down/up, `C-v`/`M-v` to page, and `C-M-<`/`C-M->` to jump to the start or end. In detail, `C-j`/`C-k` cycle sections. `M` can be Alt or Meta. Modified punctuation and `C-j` need a terminal that reports those key combinations; `g`/`G` and Tab/Shift-Tab remain available.

Closes #397


wesm:
Greetings fellow emacs user

salmonumbrella:
@wesm 🫡

PR #408, opened 2026-09-19
feat(cli): add a global inbox and standardize --all

salmonumbrella:
## What changed

- Added `kata inbox --all` to read one recipient's attention requests across active projects on the selected daemon. Results include project names and qualified issue refs in human, agent, JSON, and context output.
- Standardized cross-project scope on `--all` for `inbox`, `events`, `digest`, and `mcp serve`, matching `list`, `ready`, and `next`. Removed `--all-projects`; scripts and MCP launch configurations using that spelling must switch to `--all`.
- Kept inbox project-scoped by default, with its existing output and context truncation behavior. Global inbox reads check daemon API compatibility and fail without partial output when access is denied.
- Updated CLI help, scope errors, quickstart guidance, the agent contract, and reference docs.

## Why

An actor working across projects had to inspect each project separately to find attention requests. The global inbox gathers those requests in one command, and a consistent scope flag makes the CLI easier to use.

## Usage

```sh
kata inbox --for coordinator/teammate-1 --all
kata inbox --for coordinator/teammate-1 --all --context
kata events --all
kata digest --all --since 24h
kata mcp serve --all
```

`--for` selects an exact actor or actor/teammate address; the name above is an example. `KATA_INBOX_USER` can supply the recipient instead. `--all` reads the selected daemon only. For inbox reads, it cannot be combined with `--project` or `--workspace`.


wesm:
Pushed c7321900 to this PR. Fixed findings 1–5:

- The agent contract, README, and quickstart now keep inbox reads in the current project. `--all` is an explicit option for cross-project work, so the default workflow does not require broader access or conflict with `--workspace`.
- Inbox context now shortens long issue refs and uses the same size limit policy in both modes. A long project name no longer causes its request to disappear; the full inbox still has complete refs.
- Removed the README frontmatter and the TUI comment that incorrectly claimed the daemon lacks a cross-project endpoint.
- `digest`, `events`, `list`, `ready`, and `next` now reject `--all` with `--workspace`, matching inbox and MCP instead of silently ignoring the workspace.

I left the response checks local because the generated response types differ and a shared helper would add adapters. The health check stays, and the release upgrade note remains for release finalization.

The CLI suite, lint, vet, and docs checks passed. The full Go suite had three local setup failures: two build tests needed the locked web dependencies, and one formula check needed a Ruby version selected. All three passed on rerun after setup.

PR #409, opened 2026-09-20
feat(tui): add designated Inbox view

salmonumbrella:
## What changed

- Added an Inbox view to `kata tui`. Press `I` to see open tasks from the project designated with `role=inbox`, the same source used by the web Inbox. Enter opens a task; Esc restores the previous browsing context and filters.
- The view follows changes to the Inbox designation while the TUI is open.
- Actor requests created by `kata notify` remain in `kata inbox --for <actor>`; they are a separate inbox.

## Why

The TUI had no direct way to see the designated Inbox. You had to find that project in the projects list before reviewing captured tasks.

## Usage

Run `kata tui`, then press `I` to open Inbox. If the TUI reports that none is set, designate a project with `role=inbox` and press `I` again.

This is a separate follow-up to #408.

## Preview

TUI Inbox showing open tasks in the designated project


wesm:
Simplified Inbox so it chooses its project when you press `I`. Reopen it to pick up a changed designation. Tasks still refresh live, and Esc restores your previous view and filters. This removes the background project-switching logic and the extra handling it needed across other screens.

Also cleaned up the smaller findings:

- Removed the duplicate open-task filter; the server already filters, and the test fake now honors that request.
- Removed duplicate cache bookkeeping and renamed its counter to show that it tracks changes between project lists.
- Rewrote test comments around expected behavior and made the help label lowercase.
- Removed the committed screenshot and preview references because documentation screenshots belong in the generated assets pipeline.

Kept cancellation while Inbox is loading, so a slow response cannot replace a view you opened afterward. The pre-existing PgUp/PgDn sizing issue is unchanged.

TUI tests, focused race tests, build, vet, lint, and documentation checks pass.

